PR #63, opened 2015-01-07
changes to template structure

GLWalker:
changes to template structure to wrap each section of site for more control design wise.
Move info and error messages down into #bodyContent for better view and to avoid breaking up under the header.

//EDIT:
header.php and footer.php moved back after previous commit, leaving backwards compatibility in place. 
//EO-EDIT

HTML5 markup, we already are loading html5shiv so may as well use them, or can be covered back to standard DIV's.

### DIFF
--- a/custom.css
+++ b/custom.css
@@ -270,26 +270,26 @@ p.lead del {
 }
 
 /* Footer ------------------------------------------------------------------- */
-.footer {
+#footer {
   background: #eee;
   border-top: 2px solid #ddd;
 }
 
-.footer .footerbox {
+#footer .footerbox {
   padding: 10px 10px 0px 10px;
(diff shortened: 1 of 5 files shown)


GLWalker:
changes to template structure to wrap each section of site for more control design wise.
Move info and error messages down into #bodyContent for better view and to avoid breaking up under the header.

//EDIT:
header.php and footer.php moved back after previous commit, leaving backwards compatibility in place.
//EO-EDIT

HTML5 markup, we already are loading html5shiv so may as well use them, or can be covered back to standard DIV's.

GLWalker:
Maybe I shouldn't be pulling so much, anyhow, revert aside back to div

PR #102, opened 2015-02-08
Modularize the message stack output

sweetnessandlightcandy:
Modularize the output of the message stack


Gergely:
:ok_hand: You put the divs into the template file.

gburton:
Thank You

PR #110, opened 2015-02-11
Prettify account page

sweetnessandlightcandy:
Trial run for making the account page display a bit nicer


skleiner:
If you guys want me to test out any of these branches on my local shop let me know.  If you're looking for anything specific in the tests such as is this or that happening let me know that too. I just have gburton/osCommerce-234-bootstrap on Uni but can add others.

PR #113, opened 2015-02-13
Modularize product info page

sweetnessandlightcandy:


osc2nuke:
@mattjt83 is there any sort order and column specific setting.
cause i not seem to get it look decent!

PR #115, opened 2015-02-23
Email Templates and Email Modules

tgely:
Lets Go!


gburton:
@tgely - has this been 100% tested and you are 100% happy ?

tgely:
:100: % tested and I give support to this if you need :smiley:

skleiner:
I have this on a live shop.  The shop logo in the email does now appear in an email to a smartphone.  So that turned out OK.
On the admin side.  Tools>Email Preview I am not seeing anything in the table other than "Another Sections" at the top. In the testing phase the table was populated with email previews, etc.

skleiner:
image

tgely:
Skleiner
glob PHP Note: This function isn't available on some systems (e.g. old Sun OS). 

could you see that glob exist on your system? or DIR_FS_DOCUMENT_ROOT could be false.

kolingua:
One issue: admin/mail.php is not listed, as it doesn't contain the search text 'tep_mail(' but directly works with an email object.

skleiner:
It was the document root in configure.  I had $DOCUMENT_ROOT.  Changed to /home/username/public_html/ and that fixed it.  
Thanks

tgely:
kolingua

would you like to see in the email list or would you like as templated? 

I have no imagination what should we do with mail.php because there are a lot of conflict.  Send to "all customers" is critical if you have a lot of user and If you start a conversation with one customer the templated enviroment is unusefull. I dont want to say about replied emails which opened in another email client and not referal to oscommerce. 

The best way would be delete "all customers" from mail.php and make a special newsletter module with this option.

Any suggestion appreciated

kolingua:
Well, I presume that a conversation with one customer is more likely to take place in an email client outside osC, as incoming mails cannot be browsed in osC. And if there is one reason why shop owners might decide to answer a client's email from osC rather than from the email client, maybe it's precisely the template...

PR #117, opened 2015-02-25
Remove Old Skool Form Checks

gburton:


gburton:
Could someone test this ?  I'm 99.9 sure we dont need these for, checks as we now use the html5 required attribute.

skleiner:
I'll test in local test shop by running through the actions apparently affected by the changes.  In the next day or so.

GLWalker:
I do not think the Safari browser will use the HTML5 required attributes. Ive had issue with that before. I will test to be sure.

gburton:
Thanks Guys.  Look like partial support in Safari, with no support in iOS Safari.  http://caniuse.com/#search=required  

Do we care about Safari ?

GLWalker:
Just setup and tested in Safari on Mac osX Yosemite. It goes straight to server side checking. Safari does have its good share of users, but with server side in place no reason not to move forward.

skleiner:
On Firefox and Chrome testing affected forms..no problems.

PR #118, opened 2015-02-25
minor optimization

gadlol:
no need to call  $page_content = $oscTemplate->getContent('login'); on top.

Also there is no need for the variable $page_content . Can simply do $oscTemplate->getContent('login'); directly inside the div.


gadlol:
please close that its wrong, because the action process is called inside the module...

PR #119, opened 2015-03-01
Datepicker i18n

kolingua:
I think I have now solved the issue with the datepicker localization and autoclose function, using a newer datepicker version from https://github.com/eternicode/bootstrap-datepicker
Please check if the solution is appropriate.


kolingua:
Too many commits and unneeded merges due to my lack of experience with github, so I'll close this request and open a cleaner one.

PR #121, opened 2015-03-04
Bootstrap Settings

gburton:
Extends https://github.com/gburton/osCommerce-234-bootstrap/pull/49

Needed to allow the display to sit properly depending on the option chosen at;
admin > configuration > bootstrap setup

### DIFF
--- a/includes/footer.php
+++ b/includes/footer.php
@@ -16,14 +16,14 @@
 
 <footer>
   <div class="footer">
-    <div class="container-fluid">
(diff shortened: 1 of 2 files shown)


gburton:
I've tested, but could do with another person to have a look please.  This change allows the footer boxes and navbar to sit nicely inside the templated design.

skleiner:
Looks OK in Firefox, Chrome and IE

gburton:
Thank you Steve

PR #123, opened 2015-03-08
Ht title separator for product, category and manufacturer meta titles.

skleiner:
These are using the files from the HT reloaded add on for bootstrap, and then incorporating the concept of allowing the shop owner to choose whatever separator be it comma, dash, pipe, etc.


gburton:
Steve many thanks, these "seo" things dont (yet) exist in BS, so adding the changes at this point would break these modules...will revisit it though.

skleiner:
OK, I'll keep it in the hopper until then.

PR #124, opened 2015-03-09
Minimum Length - First Name

gburton:
Checks that the minimum length set in admin has been adhered to.

### DIFF
--- a/account_edit.php
+++ b/account_edit.php
@@ -165,7 +165,7 @@
   <div class="form-group has-feedback">
     <label for="inputFirstName" class="control-label col-sm-3"><?php echo ENTRY_FIRST_NAME; ?></label>
     <div class="col-sm-9">
-      <?php echo tep_draw_input_field('firstname', $account['customers_firstname'], 'required aria-required="true" id="inputFirstName" placeholder="' . ENTRY_FIRST_NAME . '"'); ?>
+      <?php echo tep_draw_input_field('firstname', $account['customers_firstname'], 'required aria-required="true" id="inputFirstName" minlength="' . (int)ENTRY_FIRST_NAME_MIN_LENGTH . '" placeholder="' . ENTRY_FIRST_NAME . '"'); ?>
       <?php echo FORM_REQUIRED_INPUT; ?>
     </div>
   </div>
--- a/create_account.php
+++ b/create_account.php
@@ -288,7 +288,7 @@
       <label for="inputFirstName" class="control-label col-sm-3"><?php echo ENTRY_FIRST_NAME; ?></label>
       <div class="col-sm-9">
         <?php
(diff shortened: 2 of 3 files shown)


gburton:
Should this be added ?  Would need to do the same for a lot more "min length" things.  Or should we just rip out the minimum length settings entirely.

skleiner:
Using just and initial instead of the full name has legitimate purposes.  Helping maintain a bit of anonymity for whatever reason.  For example a female may wish to just use an initial and not provide gender information for legitimate reasons.

MOPS-DanCole:
HI Gary, I just saw this and realized that you could discuss these changes here.  I think I would keep it simple and not force minimum lengths.  Required and not required goes far enough from my point of view. 

Dan

gburton:
It sounds like you both feel that this minlength should be ripped out ? @skleiner @MOPS-DanCole

skleiner:
That's correct.  Just wondering, is there a max limit?

MOPS-DanCole:
Yes I feel that way Gary...I don't see the need for a minimum length especially if we have control over whether a field is required or not.

gburton:
OK, I'm going to rip out the minlength fields.

gburton:
Said that, I do quite like the flexibility of:

minlength="' . (int)ENTRY_FIRST_NAME_MIN_LENGTH . '"

skleiner:
If minlength="' . (int)ENTRY_FIRST_NAME_MIN_LENGTH . '" would allow the shopowner to set their on min length in a config type setting then that is the best route I'd think.

gburton:
@skleiner - the configs are already in the DB (and in fact are used in post submit checking, eg:

https://github.com/gburton/Responsive-osCommerce/blob/master/create_account.php#L67-L71

What I think would be "nice" is to give an indication of that potential prior to submission, and that can be done by adding

minlength="' . (int)ENTRY_FIRST_NAME_MIN_LENGTH . '" 

to the (eg) first_name input box.

However, the html5 minlengh is not supported in mny browsers (yet), but it is accepted and coming in all browsers...

gburton:
See http://caniuse.com/#search=minlength

skleiner:
I'd forgotten about the min lengths in config, that's how long it's been since I've looked at them, and now I rethink my opinion.   If I'm setting min length of first name in config to be "1", then that probably addresses my situation.  And because some shop owner may want say..a min of 2 or 3 for their own good reason, perhaps the code should enforce that?  I can't think of an example to make that case though.   Sorry to be flipping around on this but new understanding  brings new point of view.  

but I believe I get the point now of (int)ENTRY_FIRST_NAME_MIN_LENGTH, and if that means the customer is going to see what is required of him/her, that's a good thing.

MOPS-DanCole:
I'm not having any second thoughts Gary....if we can control whether input is required for a particular field I think that should be sufficient.  Requiring some input for a particular field would ensure that at least one character would need to be entered and would address Steve's situation.  I can't think of a reason for wanting to force a minimum number of characters except for perhaps something like a password where some special handling might be required.

PR #125, opened 2015-03-09
Nav bar bug

gburton:
This is a definite bug.


gburton:
Removes an extra div that is not needed.  This div affects the display of the navigation bar, not in a particularly bad way, but it is defiinitely incorrect.

PR #126, opened 2015-03-10
SEO - Manufacturers

gburton:
Adds Metas and Description.

### DIFF
--- a/admin/includes/functions/general.php
+++ b/admin/includes/functions/general.php
@@ -1472,4 +1472,25 @@
       return is_writable($file);
     }
   }
-?>
+
+  // seo
+  function tep_get_manufacturer_description($manufacturer_id, $language_id) {
+    $manufacturer_query = tep_db_query("select manufacturers_description from manufacturers_info where manufacturers_id = '" . (int)$manufacturer_id . "' and languages_id = '" . (int)$language_id . "'");
+    $manufacturer = tep_db_fetch_array($manufacturer_query);
+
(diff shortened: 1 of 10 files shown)


gburton:
To test on an existing setup, here is the needed SQL:

ALTER TABLE manufacturersn ADD manufacturers_seo_title VARCHAR( 128 ) NULL;

ALTER TABLE manufacturers_info ADD manufacturers_description TEXT NULL, ADD manufacturers_seo_description TEXT NULL, ADD manufacturers_seo_keywords VARCHAR( 128 ) NULL;

skleiner:
Is using <br /> vs <br> a big deal in an html doctype? I see both are being used in this commit.  The <br> in the manufacturer.php file for one.

skleiner:
code removed above
br /   vs br is what I meant

gburton:
HTML5 says that it can be plain old br or br /

br is a void element, and void elements can optionally be closed with />

So, either is fine.  We should probably standardise one way or the other, but it doesn't really matter.

What do you think ?

skleiner:
If there's a trend in html markup going one way or the other. I'd say use that.  Meaning in a couple years br or br / may be being phased out....or not.
If not, and it doesn't matter go with the one that needs changed the least.
Another note, as the need for the file defines is being phased out do you plan to update that file or just wait until it's totally obsolete.
Domestic duty calls, so it'll be a couple days before I can install this on my live shop.  When I do I'll let you know how it's going. 
thanks

PR #127, opened 2015-03-11
Optimise the Gallery Layout

gburton:
Change ID to CLASS.  This allows the system to be more flexible.

### DIFF
--- a/custom.css
+++ b/custom.css
@@ -17,12 +17,11 @@
   float: right;
 }
 
-#piGal {
-  float: right;
-  width: 250px;
+.piGal {
+  width: 250px !important;
 }
 
-#piGal img {
+.piGal img {
   max-width: 250px;
   height: auto;
 }
--- a/includes/modules/header_tags/ht_product_colorbox.php
+++ b/includes/modules/header_tags/ht_product_colorbox.php
@@ -46,7 +46,7 @@
           $oscTemplate->addBlock('<script src="ext/photoset-grid/jquery.photoset-grid.min.js"></script>' . "\n", $this->group);
           $oscTemplate->addBlock('<link rel="stylesheet" href="ext/colorbox/colorbox.css" />' . "\n", 'header_tags');
           $oscTemplate->addBlock('<script src="ext/colorbox/jquery.colorbox-min.js"></script>' . "\n", $this->group);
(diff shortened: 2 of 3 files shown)


gburton:
Sets the width of the Gallery directly into the .css rather than having to amend the code of the gallery module.  Makes it more flexible and easier to manipulate.

kolingua:
I have observed something with the gallery that is not affected by this change, but I think it's worth mentioning: part of the images is clipped. Do you know why?
gallery1
gallery_2

gburton:
I have noticed this previously, and could not find the reason (admittedly, I did not look hard).  I suspect it is due to some js overriding in photosetgrid...

As for the commit, did you test it ?

kolingua:
I have seen that ext/ext/photoset-grid/jquery.photoset-grid.min.js contains the code `c=b.attr("height")*parseInt(b.css("width"),10)/b.attr("width"),e=Math.floor(.025*c);a(this).height(c-e)` that results in the `<div>` that contains each row having its height set to about a 2.5% less than it "should". It gets solved by changing it to `c=Math.ceil(b.attr("height")*parseInt(b.css("width"),10)/b.attr("width"));a(this).height(c)`. 
Apart from this, if the images in a row have different height/with ratios the "higher" ones are clipped and centered, but this is logical.

skleiner:
Installing on a live shop, and all appears to be functioning well.
Regarding what kolingua notes re cropping, I have that color script module activated but there is no cropping in my display image.

kolingua:
Well, the cropping does only affect the lowest 1/40 of the image, so if you don't have any relevant detail there you won't probably notice, usually it's just white.

kolingua:
In fact, I have now seen that the code I mentioned is commented in the full .js file (https://github.com/stylehatch/photoset-grid/blob/master/jquery.photoset-grid.js) with "Adding a buffer to shave off a few pixels in height", so I have opened an issue there asking the reason for doing so.

skleiner:
Sorry, I meant to say clipped not cropped, referring to the images you posted above.  In my live shop I don't see any cropping with 3,4 or 5 images per item.  

Regarding the clipping I'm OK with editing my shop for that but probably overall it would be best for that update  to be integrated in BS Edge or whatever we're calling it.

kolingua:
As English is not my native languages I was not really aware of the differences in the meaning of "clipping" and "cropping". The cropping of the upper and lower ends doesn't depend on how many images you have in a row, but on whether they have different heights (once adjusted to have the same width). In my second screenshot [from the demo shop](http://www.template.me.uk/2334bs3/product_info.php?products_id=28&osCsid=4b88faa248e7a271fb3c5469c0805753), as Burton has introduced a wider landscape picture, the cropping can clearly be seen in the white cellphone picture in the middle of the row.
As for your last comment, you're right, and that's why [I have reported the issue](https://github.com/stylehatch/photoset-grid/issues/50).

PR #129, opened 2015-03-13
Cleaned up deprecated PHP variables

tmccaff:


tmccaff:
Fixing some issues and deleted repository

PR #132, opened 2015-03-16
Installer

gburton:
Introduce Installer to Bootstrap


gburton:
Maybe run a couple of test installs ?  Thank You @skleiner

skleiner:
Ok..some feed back here as I go along.
First just FYI, this text is on the start up page

Welcome to osCommerce Online Merchant v2.3.4!

osCommerce Online Merchant helps you sell products worldwide with your own online store. Its Administration Tool manages products, customers, orders, newsletters, specials, and more to successfully build the success of your online business.

osCommerce has attracted a large community of store owners and developers who support each other and have provided over 7,000 free add-ons that can extend the features and potential of your online store.

skleiner:
Then the server capabilities look good

installertest

skleiner:
suggestion, during step two even thought the spinning icon should do the trick, perhaps gray out the "continue to step 3" so a trigger happy installer doesn't click too soon?  I haven't tried clicking during this process but it appears clickable

skleiner:
got to the end of the install but ran into a problem running the shop on uniserver, the folder I have this shop in is bs_installer and uniserver doesn't seem to like that, it put just bs_ in configure.php instead of the bs_installer
I will change that name to something else and resinstall to see what happens

skleiner:
Couple of things...on uniserver the configure does not set up correctly.  Maybe this is a uniserver issue but to get the shop to at least show we need this:
  define('HTTP_COOKIE_PATH', 'installtest/Responsive-osCommerce-Installer');
  define('HTTPS_COOKIE_PATH', 'installtest/Responsive-osCommerce-Installer');
  define('DIR_WS_HTTP_CATALOG', 'installtest/Responsive-osCommerce-Installer');
  define('DIR_WS_HTTPS_CATALOG', 'installtest/Responsive-osCommerce-Installer');
but instead after install it shows this:
  define('HTTP_COOKIE_PATH', '/');
  define('HTTPS_COOKIE_PATH', '/');
  define('DIR_WS_HTTP_CATALOG', '/');
  define('DIR_WS_HTTPS_CATALOG', '/'); 
which causes a fail

skleiner:
Then after getting index.php, with the above fix, there is no styling as so
installertestnostyle

gburton:
Thanks for testing.  Text "install" cannot be used in the URL, it gets stripped.

gburton:
I haven't properly hunted, but I think the "problem" is at $www_location = substr($www_location, 0, strpos($www_location, 'install')); where it looks for the first instance of "install" and anything before it becomes $www_location.  $www_location is then used to set up the configure files.

skleiner:
I had a hunch on the word "install"...I reworded things and at least on the second install I got a bit further.  I'll give this a third try later on and try to figure out why the styling isn't showing.  On the admin side, at least the standard styling is coming through.  I'll look through the template.top files.

gburton:
If you install it into (say) localhost/testing123/ (with no further subfolder) it should work.  So it ends up as localhost/testing123/index.php 

Let me know?  I've now tested multiple times and it is only problematic when the word "install" appears anywhere in the URL...

skleiner:
That did it!  The install word is verbotten in the set up process as you determined.  All's good.  Onward and upward.

gburton:
Thank You for all the testing Steve.

PR #134, opened 2015-03-16
Datepicker i18n

kolingua:
I think this solves the issue with the datepicker localization and autoclose function, using a newer datepicker version from https://github.com/eternicode/bootstrap-datepicker
I have disabled future dates for clients' birthdates as well.
Please check if the solution is appropriate, ideally using different language localizations and date formats.


kolingua:
I closed the previous pull request and opened a new one hoping to make it cleaner, but it still contains a lot of merges that don't have to do with the issue and I don't know how to avoid it. Only the two last commits are relevant.

gburton:
Thank you @kolingua - those extra commits don't matter.  Are you certain this this is tested 100% and is ready for primetime ?  I have no way to check as I do not use other languages.

kolingua:
In order to be more sure, I have just tested it again on a new installation and tried a variety of languages (just changing the definition of JQUERY_DATEPICKER_I18N_CODE in includes/languages/english.php).
It works nicely except for what seems to be a CSS issue with righ-to-left languages like Arabic or Hebrew (the datepicker spans all the way to the left border of the viewport), but it's still usable even for these languages. I'll report the issue anyway.

kolingua:
Now I'm wondering if the issue I commented is a real one. I was trying to show a right-to-left datepicker on a left-to-right page, and I have no idea of what would happen if I actually had a store which used Arabic.

gburton:
@kolingua - I'll try to run a few tests tonight...

PR #135, opened 2015-03-16
 	Complete SMTP Configuration Addon

tmccaff:


gburton:
We are not adding broken addons into the core.

PR #138, opened 2015-03-17
Replace substr with strpos as it it is faster to search in string

acidvertigo:


skleiner:
Shop owners comment here...both Jack and Roberts SEO Urls use code in the shop side html_output.php that Luca updated and Gary committed.  I use the SEO Urls in the shop so that update won't apply to my situation as the code is removed by the add on.  This is what happens when "add ons" affect core files...just the way it goes.  That's all, just commenting.
Beyond that, it's great to see Luca helping improve this software.  Much appreciated.

acidvertigo:
Thanks i'll try to update this branch when i'll have time.

acidvertigo:
There is an error in this commit.

I fixed it on this https://github.com/gburton/Responsive-osCommerce/pull/151

PR #150, opened 2015-03-18
Removal of Filename Definitions from admin

irishguy9:


gburton:
Thanks for this.  I will look at the commit in due course.

irishguy9:
I didn't realiize I added a .htaccess file and remove security dir in this commit. I thought I did that in another one. However, I think that should be done anyway. Still learning on how this is done pulling, etc

PR #154, opened 2015-03-18
Return directly last element of Cpath array instead of calculate it

acidvertigo:


acidvertigo:
Wow! You are the fastest commiter of the west!!!!

PR #157, opened 2015-03-18
Optimize for and while loop

acidvertigo:


skleiner:
Or maybe not, I just noticed this was closed.

PR #170, opened 2015-03-26
Removed some filename definitions

irishguy9:


gburton:
There are no further commits in this Repo.  
Thanks

PR #172, opened 2015-03-30
Modular product info

raiwa:
Based on: sweetnessandlightcandy:modularize-product-info-page.
Added align and margin class.
Joined review and add to cat buttons into one module.

### DIFF
--- a/custom.css
+++ b/custom.css
@@ -17,11 +17,12 @@
   float: right;
 }
 
-.piGal {
-  width: 250px !important;
+#piGal {
+  float: right;
+  width: 250px;
 }
 
-.piGal img {
+#piGal img {
   max-width: 250px;
   height: auto;
 }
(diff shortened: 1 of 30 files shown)


gburton:
We do not want options of display in the module.

BrockleyJohn:
Are you saying a breakdown to this level modularity is a bad thing? I've just done some add-on development over the top of this which has allowed me to modify the price and image displays and squirt extra stuff into product_info in the required places with no "core" code changes. Not that it's "core" but you know what I mean.

gburton:
Yes, we do not want loads of options available in the module.  Where does it stop ?  Consider that you have to account for XS, SM, MD, LG - left, right, center and so on and on and on.

For this the user will amend the **template** file to fine tune the look they need.

BrockleyJohn:
So to get away from the appearance side of things, is a product class - based approach being considered instead to give customisability without having to embed hooks at the same sort of level of detail as the modules in this branch? I was modifying the calculation of price, putting captions onto images and adding extra content stuff (eg. this product is excluded from shipping by Royal Mail). Most of that you could do by extending a core class and overriding bits and pieces.

I've working on top of this modularisation for a couple of months and it seems to hold up pretty well at all screen sizes (though most of it using desktop at different browser sizes tbh). That said, I can see it  being a total support nightmare as there's so much scope for fiddling and screwing it up. And asking, couldn't it just do....

PR #175, opened 2015-05-06
Eliminated product view count from spiders.

clustersolutions:


gburton:
Requires Testing Please

clustersolutions:
I am not too sure how to go about testing as a spider. I did test the opposite of counting only spiders and it was ok. Pls LMK...Thx!

PR #177, opened 2015-05-08
bootstrap changes applied to set password page

auzStar:
change to responsive layout, table and file names updated


gburton:
Thank You

PR #178, opened 2015-05-12
Product colorbox fix

BrockleyJohn:
Product image popus not working - the script has got .piGal instead of #piGal


gburton:
.piGal is correct

BrockleyJohn:
My apols - seem to be working with an outdated version of pi modules... must have missed a github notification somewhere

PR #186, opened 2015-06-12
Password Input Fields

gburton:
Use HTML5 type=password.
Deprecate tep_draw_password_field (shop side).


skleiner:
Thanks for continued work on Edge

PR #187, opened 2015-06-24
Update bm_card_acceptance.php

crazy-colorado:
Provide basic usage for card acceptance.


crazy-colorado:
My other suggestion would be to change the section title from "New Cards" => "Available Cards".  As this is a preference of wording.  I will leave it to the team to determine if this is to be changed.

PR #189, opened 2015-07-06
Contect modules

crazy-colorado:
Makes the Admin Content Module management more consistent by providing the Content Width description to those modules that contain the field.  Also addresses Undefined Constants on the catalog side when any of the footer or header modules are installed (this fix can be improved on depending on how the core team would like to address).

  Issue was discussed on forum http://forums.oscommerce.com/topic/407945-undefined-constant-in-cm-header-logophp/#entry1731084 .


crazy-colorado:
PS - Please ignore the Merge Pull Request in this.  Oops.

PR #192, opened 2015-07-08
Panel heading links

crazy-colorado:
Added class="panel-heading-link" to header links.  This allows site owner to easily adjust the style.


gburton:
We do not want this type of cosmetic change, thank you.

crazy-colorado:
Would adding just .panel-heading a{color:inherit} to the css be acceptable?

gburton:
That is a cosmetic decision to be made by the individual user.

PR #196, opened 2015-07-23
Schema price content

crazy-colorado:
Just a simple tweak.  According to https://schema.org/price, the currency symbol should not be included.  Also, https://developers.google.com/structured-data/rich-snippets/products, shows the currency symbol outside the value.  This just publishes simple machine-readable values alongside more human-friendly formatting


BrockleyJohn:
Shouldn't it be price including tax?

crazy-colorado:
Absolutely.  I will correct and resubmit.

PR #197, opened 2015-07-23
Schema product price content

crazy-colorado:
Just a simple tweak. According to https://schema.org/price, the currency symbol should not be included. Also, https://developers.google.com/structured-data/rich-snippets/products, shows the currency symbol outside the value. This just publishes simple machine-readable values alongside more human-friendly formatting.  Corrected prior submission to include tax.


crazy-colorado:
True.  The simplest way of doing so without having to modify core code would be preg_replace('/[^0-9.]*/','',$currencies->display_price.

gburton:
Thank You

skleiner:
After installing this update, my shop does not display the "sale" price for and item on sale.  The regular price with the strikeout show, no sale price.
saleprice

gburton:
https://github.com/gburton/Responsive-osCommerce/pull/205/files

Try that please Steve

skleiner:
OK...after this update the sale price shows again.  Thanks

PR #199, opened 2015-08-03
Speed Insights

gburton:
Makes site slightly faster to load.

### DIFF
--- a/.htaccess
+++ b/.htaccess
@@ -36,3 +36,22 @@
 
 # php_value session.use_trans_sid 0
 # php_value register_globals 1
+
+## EXPIRES CACHING ##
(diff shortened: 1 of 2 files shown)


gburton:
Should decrease the page load slightly.

skleiner:
I had different coding doing something similar for cache expiry, but if this pull request is to be merged, I'll switch.  Thanks

jwisdom:
The expiration on pdf can cause problems with invoices generated using pdf.  However, I've been using expires like this for years and it does help.  It's best for css files to have unique names when changed becaue the browser sometimes won't update unless the reload or refresh button is clicked.

PR #201, opened 2015-08-13
Checkout Shipping/Payment Address Back Button

crazy-colorado:
As discussed on forum, http://forums.oscommerce.com/topic/408143-back-button-for-checkout-shipping-address-and-checkout-payment-address .

Did not move away from the FILENAME_s defines as the rest of the page is still utilizing them.


skleiner:
If this is accepted into Edge, would it make sense to deal with all the FILENAMES in these two files at this time?

gburton:
Should we be adding in this sort of thing (or, put another way --- diverting from the official 234) ...  where does it end ?  oscMax ?

crazy-colorado:
Not sure about oscMax.  This seems to be a basic function of the site that leaves a customer in limbo, though I may be wrong.  I figured that a lot of people use this code and do not give back.  This is just my way of helping the cause.  Please feel free to use as seen fit.

As for FILENAMES, my opinion is that should be something that is done all at once rather than one file at a time.  I'll leave that to B to determine when that is to be tackled or to defer to 2.4.

PR #204, opened 2015-08-26
Manufacturer Descriptions

gburton:
Adds Description to Manufacturers Page (index.php)


gburton:
ALTER TABLE manufacturers_info ADD manufacturers_description TEXT NULL;

For those who do not want to run the full osCommerce install...

skleiner:
Will do, during next day or so. 
For the database, I will change my TEXTMEDIUM  to TEXT for consistencies sake

PR #206, opened 2015-08-28
Product holder class

crazy-colorado:
Changed the class from "thumbnail" to "productHolder equal-height".  Also
suggest changing col-sm-6 to col-sm-4.   These modifications make the two files consistent
with product_listings output


gburton:
Please not any things that are cosmetic.

PR #207, opened 2015-09-01
Glyphicon Removal

gburton:
Remove Glyphicon in favour of FontAwesome


skleiner:
OK so this is part of prep for Bootstrap 4 then.  Good. I see the drop glyphicons in that link there under big picture.  Will be working on this.  Thanks.

crazy-colorado:
I just updated my test system with this update.  I've noticed that there are entries in the bootstrap css for fa and fa-angle-****.  These entries are not used due to FA.css being loaded afterward.  Are these entries needed?

gburton:
Bootstrap File changes have been reversed.
Mistake by me to include those changes.
Thanks for the Heads Up.

crazy-colorado:
No problem.  I just also realized, it may be a good idea to move FA.css load to above the user.css.  At this point, it will override any changes we try to make to customize the cosmetics of FA

stevang:
what am i doing wrong? after changing in the main language file english.php and german.php from

`define('FORM_REQUIRED_INFORMATION', '<span class="glyphicon glyphicon-asterisk inputRequirement"></span> Required information');
define('FORM_REQUIRED_INPUT', '<span><span class="glyphicon glyphicon-asterisk form-control-feedback inputRequirement"></span></span>');`

to

`define('FORM_REQUIRED_INFORMATION', '<span class="fa fa-asterisk inputRequirement"></span> Notwendige Eingabe');
define('FORM_REQUIRED_INPUT', '<span><span class="fa fa-asterisk form-control-feedback inputRequirement"></span></span>');`

The red asterisk is not in the vertical center/middle of the input field anymore after moving to fa fa-asterisk

Any idea why that happens here?
https://www.eshopcard.de/create_account.php

Regards,
Stephan

PR #219, opened 2015-09-22
Product reviews schema

crazy-colorado:
Updates the product_reviews schema to use schema.org.  Test using my dev system and validated properly.


gburton:
Currently testing...

gburton:
Thank You

PR #220, opened 2015-10-07
Update breadcrumb.php

gadlol:
Changed ol to ul.
Remove $pos variable as it no needed since we can use $i for its purpose.
Added 2 space chars in front of <li so the generated source code to be more readable.
Added  . PHP_EOL so the generated source code to be more readable.

Note: the $separator paramater can also be removed but it also has to be removed by the parent breadcumb.php where the 
<?php echo $breadcrumb->trail(' &raquo; '); ?> 
gets called

### DIFF
--- a/includes/classes/breadcrumb.php
+++ b/includes/classes/breadcrumb.php
@@ -26,20 +26,18 @@
     }
 
     function trail($separator = NULL) {
-      $pos = 1;
-      $trail_string = '<ol  itemscope itemtype="http://schema.org/BreadcrumbList" class="breadcrumb">';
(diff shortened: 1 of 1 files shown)


gadlol:
Is there a reason for not merging that?

gburton:
It adds nothing of value.
It fixes no problems.

In other words, it is purely cosmetic.

gadlol:
Yes it adds value because you dont use the variable $pos so less memory to use.
Why use extra variables since we can code it else..

gburton:
Closed.  End.

PR #221, opened 2015-10-07
Update ht_product_colorbox.php Fixed javascript bug (Uncaught TypeError)

gadlol:
screenshot from 2015-10-07 22 05 39

Fixed bug so not to produce the error:

Uncaught TypeError: Cannot read property 'substring' of undefined.

this error exists only when we have added one image and not many.


gburton:
Thank You.

PR #222, opened 2015-10-08
Remove Grid/List buttons

gburton:
Remove Grid List Buttons if page is not in the allowed page list.

### DIFF
--- a/includes/modules/product_listing.php
+++ b/includes/modules/product_listing.php
@@ -94,7 +94,7 @@
       </div>
 
     <?php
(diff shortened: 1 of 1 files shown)


BrockleyJohn:
Fails testing on first entry in variable eg. advanced_search_result.php

I think the test should be !=== false

gburton:
@BrockleyJohn changed, could you retest please ?

BrockleyJohn:
Now performs as expected

gburton:
@BrockleyJohn thank you for testing, Eddy too.

PR #223, opened 2015-10-12
Product reviews schema

crazy-colorado:
Correct missing > on line 57 as reported by @skleiner


gburton:
Thank You @crazy-colorado

PR #225, opened 2015-10-12
Re-introduce Gallery Layout

gburton:
If you are manually updating:
1. make the code changes
2. uninstall the module from admin
3. install it again from admin


skleiner:
Testing on a live shop, after trying sever configurations of the display, all showed properly.  Tried, 122, 133, 222, 333.

gburton:
@skleiner :+1:

PR #229, opened 2015-10-26
Update application_top.php

gadlol:
There is no need to return the products_id value after we remove a product from cart.
As you can see after I removed a product from the cart the redirect url contains the products_id.

The alert message with the name of the product that has been removed, is added in the messages stact using the add_session.

A question to osc devs. Is there another way to remove product from the cart apart from the standard x button in the shopping_cart.php page?

screenshot from 2015-10-26 23 58 24


gadlol:
This only happens when you have set DISPLAY_CART to false

PR #232, opened 2015-11-18
Update product_info.php

jwisdom:
There should only be one H1 tag per page.  While looking at optimizing my site this was flagged.  So, changing the price to H2 solves this.  From what I've read it is correct to only have one H1 tag.

### DIFF
--- a/product_info.php
+++ b/product_info.php
@@ -71,7 +71,7 @@
 <div itemscope itemtype="http://schema.org/Product">
 
 <div class="page-header">
-  <h1 class="pull-right" itemprop="offers" itemscope itemtype="http://schema.org/Offer"><?php echo $products_price; ?></h1>
+  <h2 class="pull-right" itemprop="offers" itemscope itemtype="http://schema.org/Offer"><?php echo $products_price; ?></h2>
   <h1><?php echo $products_name; ?></h1>
 </div>
 


gburton:
@jwisdom https://www.youtube.com/watch?v=GIn5qJKU8VM

jwisdom:
Thanks Gary, I like that.

gburton:
After watching it ... and doing a bit more reading up in other places, I think it would be wise to change it as per your pull-req.  What the video (and html5 specs) say is OK is having multiple h1's so long as they are in different sections.  We don't use sections as of this point in time, and in any case, these two would always be in the same section...

I'll look into it a little further.  In the meantime I've re-opened the pull-req for more discussion..

skleiner:
Some while ago I changed product_info so that the title was H1 and the price H3. I made that change because when I ran various site SEO analysis (by different service / apps) and all were telling me I had too many H1 tags on the product info page.

gburton:
https://github.com/gburton/Responsive-osCommerce/pull/248

skleiner:
After updating with the proposed code, all looks good with the h1 and h2 tagging in product_reviews, product_reviews_write and product_info.

gburton:
@skleiner @jwisdom thank you chaps

https://github.com/gburton/Responsive-osCommerce/commit/70ea8a6cad6d19f2a0287b669207fd2f4fc6b765

PR #237, opened 2015-12-18
Header tags no script

crazy-colorado:
Moved the "<noscript>" block out of header_tags as this creates an html error due to a <div> located in the <head>.  Shifted it down to Navigation.


gburton:
@crazy-colorado would it be more useful to inject it into the messagestack::header (as in the messagestack header content module ?

gburton:
@crazy-colorado do you see a lot of un-styled content when first accessing the shop after this change is made?

crazy-colorado:
I have not seen any.  I'll download a recent version and see what my dev site produces and let you know.

crazy-colorado:
Nothing strange outputted.  The code reads exactly as expected.
<style> within the <head> and the <noscript> right at the top of the <body>

I do not have a lot of enabled core modules on my dev site.  No addons installed as well.

PR #240, opened 2016-01-12
Integrate Paypal App

gburton:


crazy-colorado:
Great plan.  Are you okay with updates to this code or are you going to want to keep this the same as the addon version.

crazy-colorado:
It looks like you got rid of PayFlow and the logging feature from the app and made it into a MODULE_PAYMENT.    Are there any other main changes I should be testing?

gburton:
This is http://addons.oscommerce.com/info/9184

crazy-colorado:
I was able to get this into my dev system and noticed that the Sandbox settings are no longer supported for the cm_login_paypal.  I can help with getting this configured.

Also, I attempted to branch the PayPalApp oddly it will not allow me to fork this commit.  Not sure if I am missing something.

Piernas:
I could not test it because I don't use paypal, but it will be great if it is added to oscommerce by default because of the new modular orders page. The sooner the better so new modules can use those hooks! ;)

Piernas:
Just installed it. First time I pressed on Paypal ->Start I got the message:
Warning: constant(): Couldn't find constant OSCOM_APP_PAYPAL_PS_STATUS in E:\XXXX\apps\paypal\OSCOM_PayPal.php on line 20
Along with a 'headers already sent'. It was originated at OSCOM_PayPal->hasCredentials( )
Once I go back and press again it works, so it seems to be a config variable that has not been set before visiting the page for the first time.

BrockleyJohn:
@Piernas this is a known problem, though the line number is expected to be 200...

the fix is to insert 
`    if ( !defined('OSCOM_APP_PAYPAL_' . $module . '_STATUS') ) {
        return false;
      }
`
at the top of the function hasCredentials

BrockleyJohn:
@gburton I believe this commit has some other stuff mixed in 

There are changes to admin class order - great stuff but don't think they're meant to be with paypal. Also require database changes and migration of existing data and I can't see any corresponding install changes or red flag for existing BS store owners.

gburton:
There's nothing in this commit that is not in the Paypal App (or documentation).   There are some changes to the admin order class to deal with the Paypal App (in the legacy folder or maybe the documentation)..

This commit is out of date anyway, so I'll remove the branch.

PR #252, opened 2016-02-21
Passwords

gburton:
Add autocomplete per HTML5 specs
Standardise ID=x on Password Input

### DIFF
--- a/account_password.php
+++ b/account_password.php
@@ -75,6 +75,12 @@
 
 <?php echo tep_draw_form('account_password', tep_href_link(FILENAME_ACCOUNT_PASSWORD, '', 'SSL'), 'post', 'class="form-horizontal"', true) . tep_draw_hidden_field('action', 'process'); ?>
 
+<?php
+$customer_info_query = tep_db_query("select customers_email_address from customers where customers_id = '" . (int)$customer_id . "'");
+$customer_info = tep_db_fetch_array($customer_info_query);
+echo tep_draw_hidden_field('username', $customer_info['customers_email_address'], 'readonly autocomplete="username"'); 
+?>
+    
 <div class="contentContainer">
   <p class="inputRequirement text-right"><?php echo FORM_REQUIRED_INFORMATION; ?></p>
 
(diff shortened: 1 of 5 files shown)


gburton:
This commit does two things;

1.  changes the "label" on one password field from inputNew to inputPassword
This change standardises all password input fields, which means addon makers have an easier life.

2.  adding the autocomplete 
This allows browsers/screenreaders to more usefully work out what the form is doing/needing.  Further reading at http://scripty.walterdavisstudio.com/autocomplete/

Discussion on these two changes is welcomed.

skleiner:
I installed this on a live shop. 

Signing up as a "new" customer all seems to go well, no noticeable behavior changes in the password fields.

Once signed on as a new customer, when changing the password I observe that when clicking in the existing password box, Chrome gives me a drop down asking something like "Do you want to use the password for:"...then lists several emails or usernames that I've had chrome save sign on information for.  I presume this is something specific to chrome; it's just I've not seen that before.

When logging on as an existing customer,  I see chrome is remembering my email address but not the password.  Before installing the files in this commit, the password was being remembered by Chrome.

That's it for now.

skleiner:
I noticed in the article given for reference, recommendations for changing a password:

Whenever you are changing a password, it is very important to also include the username in the same form, so that the correct account will be updated in the Keychain. If you don’t want to have the field visible, as it is here (marked with readonly so it doesn’t get changed inadvertently) you can also use a hidden field to the same effect.

Just mentioning in the event it's a potential issue.

gburton:
@skleiner - thanks for testing Steve.  It sounds to me like account_password.php should bring up a hidden field containing the accounts email_address - so that the keychain is taken care of...

The only place we actually change the password is at account_password.php as far as I recall...

gburton:
I think https://github.com/gburton/Responsive-osCommerce/commit/3fa6f3061f01291c59a6d880bccdcd6391df7f33 should take care of the keychain.  It's a hidden form field "username", readonly, with the content being the account owners email address (which is what we use for username in osc)...

skleiner:
OK, I'll update shortly and test out.  Other than what we discussed not sure what else to test as all seems in order after doing so.

PR #253, opened 2016-02-23
Modular Navigation Bar

gburton:
This allows shopowner to mix and match the Navbar as they want on a shop by shop basis...

+ Shopowner should never have to touch the main module, just turn on/off su-bmodules
- Shopowner needs to build sub-modules if they wish to do something in the Navbar that is "unusual"

Previously this was a simple "change this template" file to add/remove from Nav, but pointy-clicky seems nicer to most shopowners.  We shall see what the reaction is when it is realised how pointy-clicky leads to more complicated code to do something relatively simplistic...

### DIFF
--- a/admin/includes/languages/english/modules/cfg_modules/cfgm_navbar_modules.php
+++ b/admin/includes/languages/english/modules/cfg_modules/cfgm_navbar_modules.php
@@ -1,0 +1,13 @@
+<?php
+/*
+  $Id$
+
+  osCommerce, Open Source E-Commerce Solutions
+  http://www.oscommerce.com
+
+  Copyright (c) 2016 osCommerce
+
+  Released under the GNU General Public License
+*/
+
+define('MODULE_CFG_MODULE_CONTENT_NAVBAR_TITLE', 'Navbar Modules');
--- a/admin/includes/modules/cfg_modules/cfgm_navbar_modules.php
+++ b/admin/includes/modules/cfg_modules/cfgm_navbar_modules.php
@@ -1,0 +1,25 @@
+<?php
+/*
+  $Id$
+
+  osCommerce, Open Source E-Commerce Solutions
+  http://www.oscommerce.com
+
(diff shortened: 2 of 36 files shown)


skleiner:
Will test shortly, this is a bit of a jump for a live shop so I'll have to install uniserver and test that way.

gburton:
@skleiner thanks Steve, there's no hurry to test it.

skleiner:
Uniserver installed...a bit of testing.  All modules installed, looking good and functioning from what I see.  I'll test a bit more but I noticed one thing, doing a trial purchase, upon checkout confirmation the currency and language buttons disappeared.  Then going to my name and clicking My Account on the drop down, those reappeared.  Perhaps that is as intended.

gburton:
@skleiner thanks for testing :+1: 

When in checkout_* pages, the currency and language sub-modules should disappear...this is to stop people from changing data halfway through the checkout.

katapofatico:
:+1: I'm totally agree

skleiner:
The L for language and C for currency work. Just looking around and see fa-language and fa-money but L and C work fine.
On checkout it's a big plus that the navbar breaks out the shopping cart and checkout.  Really handy for the customer who just wants to look at the cart...or wants to go straight to checkout. The easy choices are there.
Another big plus on the drop down under my account allowing the customer quick options on some action they might want to do.

skleiner:
Checking on android smartphone..the hamburger looks well served.  And working well.

gburton:
Great, thank you for testing @skleiner - appreciated.

@Kymation did you find the problem you mentioned at https://github.com/gburton/Responsive-osCommerce/pull/253#commitcomment-16250727

Kymation:
I never found the bug; I just imported the code again and it went away. Probably something got corrupted, but I couldn't see it and PHP refused to tell me where it was.

gburton:
@Kymation - thank you for testing, appreciated :+1:

PR #255, opened 2016-02-25
Template Top - iOS bug

gburton:


gburton:
More reading:
http://v4-alpha.getbootstrap.com/getting-started/introduction/#responsive-meta-tag
http://kihlstrom.com/2015/shrink-to-fit-no-fixes-zoom-problem-in-ios-9/

PR #259, opened 2016-02-27
Basic support for orders language

Piernas:
Adds support for identifying the language used when an order was placed. Will allow to send order updates or emails in the customer's language.

### DIFF
--- a/checkout_process.php
+++ b/checkout_process.php
@@ -118,7 +118,9 @@
                           'date_purchased' => 'now()', 
                           'orders_status' => $order->info['order_status'], 
                           'currency' => $order->info['currency'], 
-                          'currency_value' => $order->info['currency_value']);
+                          'currency_value' => $order->info['currency_value'],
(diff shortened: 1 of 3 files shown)


gburton:
When you have the full changes done and ready for testing, please push it.  Thanks.

Piernas:
I think the only other change needed involves removing strings like EMAIL_TEXT_SUBJECT from english.php (or other language files) and adding them to another language file that loads from orders.php when an email is being sent, instead of the admin's language strings. 
I'm thinking on creating a new file (say includes/languages/english/email_constants.php) that contains those constants, but there will be no corresponding file in admin.  I'll do it and test ASAP. Do you think this is the right way to go? I'm not sure if there should be loose files (not corresponding to an identically named file) in languages directory.

Piernas:
Sorry for my mistake, those constants are in languages/english/orders.php on in the main language file!

PR #261, opened 2016-02-28
Solve basename conflict

gburton:
basename(x) should not be used as this creates a problem for any files outside of the root of the shop, eg those customer facing files in /ext/


gburton:
Anyone testing this should see absolutely no changes to links in the currency/language dropdowns...everything should work 100% as normal.

PR #281, opened 2016-03-06
Remove unused global variables

acidvertigo:


gburton:
What is the benefit of removing?

katapofatico:
Clearing, avoid conffusion, reducing reading effort and avoid thinking on "what's the porpouse of this variable? Can I deleted it? Perhaps it imply a dependency because it's used in another included file?, Is it relevant to my coding porpouse?". Personally, I always keep the code as simple as possible. The rest is garbage, noise.
:+1:

katapofatico:
Two answers from http://programmers.stackexchange.com:
- [Delete unused source code from the repository or keep an archive folder?](http://programmers.stackexchange.com/a/289908/44844)
- [Should I remove unreferenced code?](http://programmers.stackexchange.com/a/103032/44844)

gburton:
There is so much defunct and old unused code in osCommerce that it is not worth trying to fix these little things at this point.  It is more important to get the important things done first.

Thoughts ?

skleiner:
As a shop owner who liked to maintain his shop coding (with help from experts) I favor cleaning out the unused code. It's like remodeling your home, every once in a while you have to stop the main project and sweep things up so your main goal can proceed unimpeded.  I'd say keep focus on building main code as has been done, but remove unnecessary snippets, etc. as they become known.

MOPS-DanCole:
I would agree with Steve on this....I wouldn't spend a lot of time on it but as I came across unused code I'd be inclined to remove it so I wouldn't have to think about it or revisit it again.

gburton:
There are more important things to do in the core code ... than this type of cleanup.  
Perhaps we can set aside a "bug week" for this sort of thing in the future.

PR #282, opened 2016-03-10
padding fix for non text links in navbar for XS and SM viewport

Tsimi78:


gburton:
Thank you @Tsimi78 
Thank you (as always) for testing @skleiner

PR #283, opened 2016-03-11
GTIN

gburton:
GTIN
aka UPC, EAN, ISBN, ISMN etc etc

Has to be stored as exactly 14 characters per GTIN rules.  
This means if you use a 13 character GTIN it will be padded with an extra zero.

### DIFF
--- a/admin/categories.php
+++ b/admin/categories.php
@@ -226,7 +226,9 @@
                                 'products_status' => tep_db_prepare_input($HTTP_POST_VARS['products_status']),
                                 'products_tax_class_id' => tep_db_prepare_input($HTTP_POST_VARS['products_tax_class_id']),
                                 'manufacturers_id' => (int)tep_db_prepare_input($HTTP_POST_VARS['manufacturers_id']));
-        $sql_data_array['products_gtin'] = str_pad(tep_db_prepare_input($HTTP_POST_VARS['products_gtin']), 14, '0', STR_PAD_LEFT);
+        if (tep_not_null($HTTP_POST_VARS['products_gtin'])) {
+          $sql_data_array['products_gtin'] = str_pad(tep_db_prepare_input($HTTP_POST_VARS['products_gtin']), 14, '0', STR_PAD_LEFT);
+        }
 
         $products_image = new upload('products_image');
         $products_image->set_destination(DIR_FS_CATALOG_IMAGES);
--- a/admin/categories.php
+++ b/admin/categories.php
@@ -226,6 +226,7 @@
                                 'products_status' => tep_db_prepare_input($HTTP_POST_VARS['products_status']),
                                 'products_tax_class_id' => tep_db_prepare_input($HTTP_POST_VARS['products_tax_class_id']),
                                 'manufacturers_id' => (int)tep_db_prepare_input($HTTP_POST_VARS['manufacturers_id']));
+        $sql_data_array['products_gtin'] = str_pad(tep_db_prepare_input($HTTP_POST_VARS['products_gtin']), 14, '0', STR_PAD_LEFT);
 
         $products_image = new upload('products_image');
         $products_image->set_destination(DIR_FS_CATALOG_IMAGES);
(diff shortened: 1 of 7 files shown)


gburton:
Sorry @skleiner Steve I must have missed your comment/question.  Yes, that would be fine for testing it.  Mikepo at the oscforum has also tested and it appears OK. http://forums.oscommerce.com/topic/409139-ean-gtin/

skleiner:
No problem, I'll probably install this over the weekend on the live shop.  Re the issue of padding with zeros if the GTIN is shorter than 14 digits...I saw the fix in the forums for that and will do that as well.

skleiner:
@gburton Gary, I have this installed and looks to be working fine.  Couple of things.  Entering a test GTIN (UPC for me) the observed result was good.  However when deleting that test UPC the box remained active with all zeros showing. I had to go back into the database to set that field value to NULL to get the box to disappear.
So FYI on that.
I also incorporated the tip from the forums to only show 12 digits, works well.
For the sake of users hereafter and to keep the same question from coming up again and again in the forum perhaps some sort of self help info in cases such as this would be beneficial.  Perhaps right there when the module is installed included in the info box to the right that has "Shows the GTIN (UPC, EAN) Number on the Product Info Page.....".

skleiner:
Gary, update on this.  What I am finding is that when I edit products for other reasons than to enter a GTIN, the GTIN will be updated to all zeros upon update. Which then shows on shopside as a UPC with the value of say.."000000000000".  I've been correcting that via phpMyAdmin so nothing shows if there is no GTIN.

gburton:
@skleiner does it go to 000000000000 even if the GTIN already has a value ?  Or only if the GTIN is blank (null) ?

skleiner:
@gburton What I am seeing is:
1) as long has nothing has been done administratively to a product admin side, all's good and nothing shows as far as GTIN shop side. As it should be I'd think.
2) if entering a GTIN deliberately, and saving..it shows properly shop side as per the module intends.
3) if I remove the GTIN, or simply work on a product and save while in admin, the null data for the GTIN is changed to all zeros and those zeros show shop side as the GTIN. Probably something we don't want.
No. 3 can be corrected in phpMyAdmin but that's a bit tedious and would have to be done every time a product is worked on in admin.

gburton:
@skleiner Thanks Steve, I'll look into it.

MOPS-DanCole:
Gary this might be one of those situations where we want to keep it simple and just let the shopowners fill in what they want..I know the GTIN site says to stuff it with zeros where it is shorter than 14 characters but I can't help but think that is just as guidance for those who are creating the GTINs in the first place. We use UPCs which are 12 characters and while stuffing it with leading 0s won't hurt it doesn't help in any way and indeed those using EAN probably would feel the same.  I think the important thing is having the field long enough to allow for use of the GTIN where it is needed or useful.  Let the shopowner decide that.

gburton:
@MOPS-DanCole - totally get it, leaving it for the shopowner to decide is an option, but leaving it like that opens the door to problems with Microdata.  I'd prefer to store it 14 and output it gtin14, and then allow individual shopowners to do as they want on the output side.  An idea for that might be to put some options in the modular file:  output 14, output 13, output 8 - and so on.  Thoughts?

MOPS-DanCole:
I hadn't thought about the Microdata aspect.   If it doesn't get to complicated and the table can be created from within the module it would be neat to allow the shop owner to decide if they wanted to use GTIN in 8 12, 13 or 14....once selected the table could be created and the appropriate Microdata GTIN used. Padding I assume would then become a none issue. I see there is a microdata GTIN-8, GTIN-12, GTIN-13 and GTIN-14 available for use.

gburton:
@skleiner @MOPS-DanCole any last thoughts before I "hit the button" to merge it?

skleiner:
Nothing I can think of. Thanks.

MOPS-DanCole:
I haven't tested it Gary and likely won't get a chance until tomorrow or Friday....I'm sure it's okay (Steve seems to have a sharp eye) but if you want to wait I'd be happy to test it for you in the next day or two.

gburton:
Thanks chaps.  I'll merge it now, and we can keep an eye on it.  :+1:

skleiner:
@gburton Gary, I'm working on another commit, but noticed some code in my files from when we're discussing limiting gtin to alpha only.  I don't see that in this commit. Just FYI in case it was supposed to be there.

In admin/categories.php  'pattern="\d+"'

`<td class="main"><?php echo tep_draw_separator('pixel_trans.gif', '24', '15') . '&nbsp;' . tep_draw_input_field('products_gtin', $pInfo->products_gtin, 'pattern="\d+"'); ?></td>`

stevang:
is there a way to have different type of gtin in the shop currently? as i understand it you have to decide how you gonna display the gtins for all your products as it is now, right? gtin13 gtin14 etc

gburton:
Correct.  You can make some sort of interface to allow yourself to say "13" or "14" or whatever.  Then amend the module to reflect that choice.

PR #287, opened 2016-03-14
Install modules

gburton:
Install some modules so that the shop looks a little more shoplike installed out of the box.


MOPS-DanCole:
Great idea Gary...

katapofatico:
:+1:

PR #294, opened 2016-03-17
Allow Admin to search products by model number or GTIN number.

Tsimi78:
Now that we have GTIN numbers why not allow the admin to search products by it. 
And while we're at it why not also add a search by model number feature.

Thoughts?


skleiner:
Seems the concept would be beneficial in shops with a large inventory where searching by product name or parts thereof would yield many results.

BrockleyJohn:
Model is more generally useful, I think. GTIN is far too long to type in manually and if you've got barcode scanners you're probably not using osc for stock control! That said, you could type in a clump of digits from the middle.

Apart from the need to publish them for search engines, GTINs are mainly used:
Goods inwards (delivery receiving)
Stock check
Sales (creating an order)

My view is - store-specific customisation. I'll bet lots of people customise this search anyway. I've done it for several clients with quite different requirements.

Tsimi78:
@BrockleyJohn If you have done this mod for various shops already why not have it build in and save you the time from now on? Not that it takes much time to insert the code but something less on the "to do" list. ;-)
Of course the search works if you enter partial GTIN or model numbers. When I had a shop I used different product names then my supplier and the only way to find the right product quickly was to search by model number. Maybe someone might find it useful to search by GTIN? I never used such numbers since osC didn't had such feature.
@burt what do you think?

gburton:
@Tsimi78 it just needs a couple of testers to say all is well ...

BrockleyJohn:
setting up a new install for testing... [the other 8 need to stay where they are!]

BrockleyJohn:
Tested partial & full matches on gtin & model, as well as regression against name. 

In doing so, also tested install and admin of pr/283 with 13-char EAN format and 14-char.

No issues found.

skleiner:
Question: In this case I can work with the raw file view to create a testing file, but for more complicated pull requests from** other contributors** is there a way to **pull that person's pull request** into a branch on my local repo for gburton/Responsive-osCommerce for testing? In this case pull Lambros' 294 --> skleiner/Responsive-osCommerce into branch "testing" for example.
Basis for question is: I'd prefer to stay working within gburton/Responsive-osCommerce and my own local branch rather than create an entire repo for each person who has a pull request up for consideration.  **Using Sourcetree for all this.**

skleiner:
Testing search for model and gtin in admin, entering partial strings....a all good

BrockleyJohn:
@skleiner you can pull the pull request branch straight into your local sourcetree repo... if you want to pull a branch from another fork of resp-osc that's not in gary's repo you may have to add the other fork as another remote. It doesn't have to hit your github repo unless you push it. IIRC this was one of the things we experimented with in Dan Cole's "lets all learn sourcetree together" thread

skleiner:
@BrockleyJohn Thanks after my post I figured out how to do this.  In Github made a branch "lambros" for my responsive repo fork, pulled Lambros repo into that.  Then in Sourcetree, added my repo as as remote,  then from there pulled the  "Lambros" branch into my working copy right there in Sourcetree, now giving me clean and quick access to his commit data.

gburton:
Is this needed/useful for shopowners.  Do any shopowners need to search via GTIN ?  
How do they know a particular products GTIN (in order to start a search?)

MOPS-DanCole:
Gary not to my way of thinking....we use a model numbering system that we created and can easily figure out whenever we're looking for something....seems to me that this is the sort of thing a shop owner might add to suit his own individual circumstances and not something that needs to be added to the core.

skleiner:
Now that I have this installed on a live shop I'll keep it there and could see myself searching by model number or a part thereof.
My shops are small, so I'm not sure I'll search by GTIN.

MOPS-DanCole:
Gary if you're asking about the model number too, I would definitely use that...I have it added to the search function on my site and have been using it frequently for years.  So I would vote yes for model number and no or don't care for GTIN.

gburton:
Thank you @Tsimi78 
Thank you @skleiner @BrockleyJohn @MOPS-DanCole for testing and commenting.

PR #297, opened 2016-03-20
Update jQuery to v2.2.2

gburton:


skleiner:
Testing on live shop. Preliminary results with no observed issues. Site browsing, log in, test purchase, log off.
If there's a specific routine that uses the new jquery that needs tested please advise.

gburton:
@skleiner thanks (as always!) for testing.  No particular testing procedure other than general use.  I've been using 2.2.2 on a few test shops for a week with no problems seen...

Tsimi78:
Also quickly installed and tested on my local (XAMPP) test shop. Looks good thus far.
No conflicts or hiccups on both sides (catalog & admin area)

Piernas:
Tested on live test shop. No issues found.

gburton:
Thank you all

PR #303, opened 2016-03-22
Enforce Digits Only for GTIN

gburton:


skleiner:
@gburton Got a pop up box saying **please match the requested format**.  Works for me since I know what's requested.  Can the pop up be tweaked to say **Please Enter Numeric Only**.  My point here is I think you probably need to help those shop owners using _responsive_ who just don't know and/or care about technical issues such as formatting, etc, as simple as they may be.  Or less tactfully...dumb it down as much as possible for them.

BrockleyJohn:
Would there be any point in checking the number of digits entered - either that it matches the catalog setting or at least that it's not longer than the field? Testing showed that extra digits fall off the end so I infer it's held as a string.

gburton:
Where does it stop?  Let's let the shopowner have some responsibility for their (in)actions.

BrockleyJohn:
I agree - you're the one started writing validation :P 

Seriously, though, more than 14 chars is almost certainly a typo or a double-scan and truncating it silently means no one will ever know.

gburton:
Yeah, maybe having validation is not needed.  Shopowners need to take responsibility for their actions.

PR #307, opened 2016-03-22
Orders class - added order_id, order_status_id and order_status_name

Piernas:
- Added **order_id** to the info set of properties.
- Added **order_status_id** to the info set of properties.
- Added **order_status_name** to the info set of properties. added an
**order_status_query** and global variable **$language** to the admin side **needed ** to
get order_status_name in the admin language.
Tested and working on my side. Made some modifications to admin/orders.php and account_history_info.php on my test site to retrieve the properties and check if it works fine. Every property can be fetched and used. 
All in one single commit.

### DIFF
--- a/admin/includes/classes/order.php
+++ b/admin/includes/classes/order.php
@@ -24,6 +24,8 @@
     }
 
     function query($order_id) {
+      global $languages_id;
+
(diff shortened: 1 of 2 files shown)


gburton:
OK, so why do we need this data as part of the orders class ?

Piernas:
Order id will avoid the need, for example, of passing $_GET, $_POST or $oID to other classes or modules, and get order id whenever the class has been initialized. Code will be much cleaner and easy to use.

order_status_id and order_status_name will be also useful whenever we need to display order status name or id without doing extra queries in the code. Also makes shop and admin orders more similar so they can be used in the same way.

gburton:
OK, that gives info to anyone who'd like to test it.  Thanks!

katapofatico:
...an very useful example to me, extending @Piernas comment: On a extended using of order class (not only to encapsulate $_GET orders_id) we can access to orders_id if we are looping over an array of orders, for example:

```
public static function getSomeCriteriaOrders($criteria) {
   // query WHERE $criteria
  // ...
  while ($orderArray = tep_db_fetch_array($someCriteriaOrdersQuery)) {
       $orders[] = new order($orderArray['orders_id']);
  }
return $orders;
}

$someCriteriaOrders = getSomeCriteriaOrders($criteria);
foreach ($someCriteriaOrders as $order) {
    // I can access to $order->info['orders_id'] and $order->info['orders_status_id'], great!;
}
```

PR #312, opened 2016-03-24
Testimonials

gburton:
Allow shopowners to add Testimonials

Testimonials are different to reviews, as reviews are per product, whereas a Testimonial is about the shop or the shopowner.

### DIFF
--- a/admin/includes/boxes/catalog_testimonials.php
+++ b/admin/includes/boxes/catalog_testimonials.php
@@ -1,0 +1,22 @@
+<?php
+/*
+  $Id$
+
+  osCommerce, Open Source E-Commerce Solutions
+  http://www.oscommerce.com
+
+  Copyright (c) 2010 osCommerce
+
+  Released under the GNU General Public License
+*/
+
+  foreach ( $cl_box_groups as &$group ) {
+    if ( $group['heading'] == BOX_HEADING_CATALOG ) {
+      $group['apps'][] = array('code' => 'testimonials.php',
+                               'title' => MODULES_ADMIN_MENU_CATALOG_TESTIMONIALS,
+                               'link' => tep_href_link('testimonials.php'));
+
+      break;
+    }
+  }
+  
--- a/admin/includes/languages/english/modules/boxes/catalog_testimonials.php
+++ b/admin/includes/languages/english/modules/boxes/catalog_testimonials.php
@@ -1,0 +1,13 @@
+<?php
+/*
+  $Id$
+
+  osCommerce, Open Source E-Commerce Solutions
+  http://www.oscommerce.com
+
+  Copyright (c) 2010 osCommerce
+
+  Released under the GNU General Public License
+*/
+
+  define('MODULES_ADMIN_MENU_CATALOG_TESTIMONIALS', 'Testimonials');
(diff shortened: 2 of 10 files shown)


MOPS-DanCole:
Are you looking to have this tested Gary?

gburton:
@MOPS-DanCole Yes please.  The necessary SQL is at the end of the install file (in the commit).

skleiner:
Just working on testing and notice that because I have the **header tags reloaded add on** installed, I am going to have to do something with the admin/reviews.php (and language), as well as how that add on modified the database before I get started. Goal is to keep ability to enter reviews via admin.  So FYI to anyone who may be in the same situation and want to do some testing.

MOPS-DanCole:
Just took a run at this but ran into issues when I went to add a testimonial....I don't see any action buttons on the testimonial.php page.  Did I miss a step?

skleiner:
@MOPS-DanCole Try this around line 183"

`<td colspan="2" class="smallText" align="right"><?php echo tep_draw_button(IMAGE_BUTTON_ADD_TESTIMONIAL, 'triangle-1-e', tep_href_link('testimonials.php', 'action=new')); ?></td>`

MOPS-DanCole:
Thanks Steve...I assume it needs a few other buttons too....if I get a few minutes I'll play around with it tomorrow.

MOPS-DanCole:
Gary I installed the updated files and it now seems to work as you'd expect...I couldn't see any issues at all but I did wonder if you had anything planned to display the testimonial page on the shop side?   I added it as a link on the footer area (as a hack) but I'm thinking you probably have something more elegant in mind.  The other issue that occurred to me is that I will now need to move all my customer site reviews out of the review table into the testimonial table.   The good news is that is probably an issue that is unique to me.

skleiner:
@gburton Installed and tested latest (with button).  All's good.

gburton:
@MOPS-DanCole maybe a navbar module ?  The reviews has one, so why not the testimonials?  That might be nice.

Or added as a link in the information box in footer or side?

MOPS-DanCole:
I added mine to the information box in the footer but the navbar is a good idea too.

gburton:
Navbar Link: https://github.com/gburton/Responsive-osCommerce/pull/312/commits/41025c36c67f56bcd0ad2ec6d266e1294c00a5ec

gburton:
Thank you for Testing chaps as always. @skleiner  @MOPS-DanCole

pitoy:
Helo @gburton 
pls if you don't mind Iam commenting this Testimonials.. just for the info I run the Responsive master download May 9, 2016, using PHP 5.3.29 and SQL 5.5.46.
I have a problem when testing the testimonials at more than one language shop.. at admin site there is only one fill in box for testimonials and it will depend which language when entering the admin site.
For example if I fill in a testimonials in english language admin site option... so at the front shop site.. the testimonials only show in the english language option... other language will display "there is no testimonials".
In my opinion it will be better if there is multi language option to fill in for testimonials (like when I fill the product description and product catalog).
best rgds,

gburton:
My opinion is that Shopowner would get a testimonial in the customers language only...
And as per EU Law he would then post that testimonial as is, with no changes...

PR #316, opened 2016-03-27
Seo products

gburton:
This give the Products the ability to have Meta Description and (optionally as in terms of SEO they are fairly useless) Meta Keywords.

The Meta Keywords functionality can optionally be set up as an Internal Search Engine.


skleiner:
@gburton Setting this up for testing but need to rework some code before I start because I have HT reloaded installed. On an aside, if there's no future plans for products_mini_descriptions from HT reloaded, I'm just go to remove that all together

gburton:
@skleiner - as a shopowner do you use the products_mini_descriptions for anything?  The feedback I got from my guys who I installed this for;  

products_mini_descriptions - never used
seo_title (for products, categories, manufacturers) - never used

Thoughts ?

skleiner:
@gburton Never used it, I only let this in there because of some possible future use.  I was thinking maybe someday used in product listing grid view, but for me not that big a deal anyway. May be for other shop owners though it could be....I think I'll remove it and if used someday, put it back in.

MOPS-DanCole:
Gary....just so you know your work isn't for not....I use the category titles....I'm finding that I can write some more meaningful category Titles for SEO purposes and keep a fairly brief category name for the drop down list or boxes within the shop.  I like that feature and I love the category descriptions.

gburton:
@skleiner @MOPS-DanCole thank chaps.  
The "problem" with the seo_title stuff is it's quite invasive.  
Thoughts?

I'm dumping "products_mini_description" for now.

skleiner:
I've removed (hopefully) all "products_mini_description" code. I'll give things a day or so to see if I get any surprises, then if not work on the SEO products.

Regarding SEO Titles, from what I see they offer some over and above SEO capabilities.  The SERP title for example can be tweaked for more detail.  I'd recommend making them core for the sake of all around SEO in responsive.

MOPS-DanCole:
I think the SEO Titles should be included in the core as well...to me its a critical component of SEO that should be used by all shop owners.

gburton:
Under consideration..thanks chaps

DougieWalker:
Hi there
never used mini description,
but always use 
seo_title and Title

skleiner:
@gburton Testing this update, no issues observed after testing with a product and category update process.
Noting different approach from HT Reloaded in handling $sql_data_array and $parameters..just and observation there.  I let SEO product and title seo intact for reasons above.

gburton:
Slightly different approach, makes it a little cleaner (in my opinion)...

Sounds like seo_title needs to be coded in.  Is this for all?  Categories & Manufacturers & Products?

skleiner:
I'd say so. Regarding Mfg title seo...That's something I've not had installed, not sure if I missed it in HT reloaded or what.  That being said, I can see a use for that as well.

gburton:
It might not have had it, I don't recall...

I'll open another issue as there needs to be a bit of discuss about seo_title ..

skleiner:
@gburton Working on this. Do we need a language file for the module?

gburton:
@skleiner as always thanks for testing Steve.  No language file for the ht_product_title, as it will use the already existing one...

skleiner:
@gburton Preliminary testing, all's good...nice.  Letting the shopowner decide on use or not of SEO title and breadcrumbs is very flexible.

Suggestion, a while back I think you provided me with coding for the previous module version that would allow the shop owner to set their own separator via the language file. Mine shows:
`	  define('MODULE_HEADER_TAGS_PRODUCT_TITLE_TITLE', 'Product Title');
	  define('MODULE_HEADER_TAGS_PRODUCT_TITLE_DESCRIPTION', 'Add the title of the current product to the page title');
	  define('MODULE_HEADER_TAGS_PRODUCTS_SEO_SEPARATOR', ' | ');
.

I've modified the module file in this pending update already to allow it to do that'

gburton:
@skleiner - thank you for testing Steve

PR #317, opened 2016-03-28
php 7 admin/classes changes - class name method is no longer constructor

BrockleyJohn:
### General approach to changes:
Rename constructor from class name function to __construct.
Files where approach was varied:
- object_info.php
- payment_module_info.php
- rss.php
- shopping_cart.php
- table_block.php

### General testing approach:
- before change, set to php7 and confirm deprecated notice appears
- confirm change removes notice
- check page behaviour before and after change is the same
- if it couldn't be executed by running core code, I didn't change it

action_recorder.php
- extends [catalog/] actionRecorder
- dynamically instantiates modules but only method call is setIdentifier
- used once in admin only to instantiate, no use of class name method
- rename constructor
- test in login.php

box.php
- extends tableBlock
- no instances of ->box found in admin
- rename constructor
- tested in banner manager

cfg_modules.php
- no direct calls of class name method found in admin
- rename constructor
- class dynamically instantiates modules - but doesn't dynamically call class name methods
- test in module admin

currencies.php
- no direct calls of class name method found in admin
- rename constructor
- test in orders and currencies

email.php
- no direct calls of class name method found in admin
- rename constructor
- test in mail

language.php
- no direct calls of class name method found in admin
- rename constructor
- test by setting get language

logger.php
- no direct calls of class name method found in admin
- rename constructor
- test by turning on logging

message_stack.php
- extends tableBlock
- no instances of ->messageStack found in admin
- rename constructor
- tested in banner manager

mime.php
- no direct calls of class name method found in admin
- rename constructor
- test in mail (instantiated for any type of email content sent)

object_info.php
- class name method is both setter and resetter
- several instances of ->objectInfo found in admin
- __construct added calling existing constructor
- tested in banner manager

order.php
- no direct calls of class name method found in admin
- rename constructor
- test in invoice, orders and packing slip

passwordhash.php
- no instances of ->PasswordHash found in admin
- constructor renamed
- test in login (for tep_validate_password) and administrators (for tep_encrypt_password), including with htaccess

payment_module_info.php
- can't find any use of this file
- class name different to file name so unlikely to be included/instantiated dynamically
- no change to contents but could remove file (??)
- tested by visiting payment modules (installed and to install), orders

phplot.php
- no instances of ->PHPlot found in admin
- constructor renamed
- tested in countries

rss.php
- no constructor defined
- no change
- test in dashboard/latest news + latest addons

shopping_cart.php
- constructor only resets object, designed to be called as a reset method too
- but it isn't!!
- empty __construct added; existing constructor left (just in case)
- tested in whos_online

split_page_results.php
- no instances of ->splitPageResults found in admin
- constructor renamed
- tested in countries

table_block.php
- never instantiated directly, only via child classes
- so doesn't need constructor, which has no means of testing
- class name method called by children so can't be renamed
- no change
- [maybe add __construct calling class name method in case instantiated in addons?]
- testing by watching for deprecated notice for this module while testing others throughout admin

upload.php
- no instances of ->upload found in admin
- constructor renamed
- tested in banner manager

### DIFF
--- a/admin/includes/classes/action_recorder.php
+++ b/admin/includes/classes/action_recorder.php
@@ -13,7 +13,7 @@
   require(DIR_FS_CATALOG . 'includes/classes/action_recorder.php');
 
   class actionRecorderAdmin extends actionRecorder {
-    function actionRecorderAdmin($module, $user_id = null, $user_name = null) {
+    function __construct($module, $user_id = null, $user_name = null) {
       global $language, $PHP_SELF;
 
       $module = tep_sanitize_string(str_replace(' ', '', $module));
--- a/admin/includes/classes/box.php
+++ b/admin/includes/classes/box.php
@@ -24,10 +24,10 @@
 */
 
   class box extends tableBlock {
-    function box() {
+		function __construct() {
       $this->heading = array();
       $this->contents = array();
-    }
+    } 
 
     function infoBox($heading, $contents) {
       $this->table_row_parameters = 'class="infoBoxHeading"';
--- a/admin/includes/classes/cfg_modules.php
+++ b/admin/includes/classes/cfg_modules.php
@@ -13,7 +13,7 @@
   class cfg_modules {
     var $_modules = array();
 
-    function cfg_modules() {
+    function __construct() {
       global $PHP_SELF, $language;
 
       $file_extension = substr($PHP_SELF, strrpos($PHP_SELF, '.'));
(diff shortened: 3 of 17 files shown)


gburton:
Thank You @BrockleyJohn

skleiner:
Looks like my uniserver will handle php 7 so if plans are to roll this into edge, I can test once I get things set up assuming it let's me set up to show errors.  If so it'll be a bit to allow set up in uniserver and take care of other things going on.

BrockleyJohn:
Fix for tabBlock pushed - thanks @raiwa for testing

PR #318, opened 2016-03-29
User.css header tags module

Piernas:
This enables the user to give priority to the file and allows other modules to place their CSS below or above user.css.
This also allows to disable the module so if you don't use it the empty file is not even requested to the server. I think it makes sense.


BrockleyJohn:
I think you need to rebase your master branch. You should be pulling Gary's commits not merging your own.

BrockleyJohn:
Do you have an example of a module that was adding a stylesheet after user.css to try this out on?

Piernas:
@BrockleyJohn I still didn't found how rebase works. I tried but made a big mess with the commits. Will give it another try.

Sorry I don't have any module that does this. For me the benefit is allowing to skip the file load if it's empty, and it also allows sort ordering it in header if needed.

BrockleyJohn:
I was using the term rebase loosely! It's actually a hard reset you need to do... see [http://stackoverflow.com/questions/29049650/github-fork-your-branch-is-5-commits-ahead-how-to-clean-this-without-pushing]

Piernas:
Thanks @BrockleyJohn !  I'll test it. It's annoying to see all those commits and I've tried several methods. One of them worked but I cannot find it anymore. I'll test that link.

gburton:
I really dislike the idea of a header_tags user.css, sorry.

Piernas:
I'm just curious, why?

gburton:
The future of .css is loading css **where it is needed**.  

So, if a module needs some css, load it by the module.  
Footer needs css, load at footer and so on.

This is something under discussion now..

frankludriks:
Open Chrome, then open a product info page. Press Ctrl + Shift + I for developer tools, click on the Audits tab, then Audit Present State, then Run.

Look at 'remove unused CSS rules'

How many do you have?

I have 3,279 (!)

gburton:
@frankludriks a typical shopowner needs a few more than is presented in a out of the box install (so that he/she can make his shop look prettier, eg a theme).  You'll find loads of unused .css rules due to the Bootstrap css.  I believe in bootsrap 4 they are thinning it down.

PR #325, opened 2016-04-05
SEO: Manufacturer Title

gburton:
Admin, Shop, Installer.

Allows shopowner to over-ride Manufacturer Name with a more precise SEO title.

If updating a live (or test) shop, run SQL:
ALTER TABLE manufacturers_info ADD manufacturers_seo_title VARCHAR(128) NULL ;
Also turn off the Manufacturer Title HT module, then turn it back on again.

### DIFF
--- a/admin/includes/functions/general.php
+++ b/admin/includes/functions/general.php
@@ -1525,6 +1525,12 @@
     return $manufacturer['manufacturers_seo_keywords'];
   }
   
+  function tep_get_manufacturer_seo_title($manufacturer_id, $language_id) {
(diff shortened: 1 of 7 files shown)


skleiner:
All's good here admin and shop side.
This may be particular to my shop, or others that have long shop names, but to benefit by adding keywords to the title I had to disable the shop name showing up in the Mfg Title.  My shop name is 24 characters and that eats up so much of what the "recommended" title string length is that if I add much to the SEO title it goes over.
` $oscTemplate->setTitle($manufacturers['manufacturers_seo_title']/* . MODULE_HEADER_TAGS_MANUFACTURER_SEO_SEPARATOR . $oscTemplate->getTitle()*/);`
Just an FYI on that in the event someone else has this issue.

PR #332, opened 2016-04-12
Shopping cart update

MOPS-DanCole:
Here are those updates Gary.


BrockleyJohn:
@skleiner did you not get deprecated notices?
@MOPS-DanCole can we have php7 constructors, please?

In other words, instead of
`class myClass {
  function myClass() {`
can we have
`class myClass {
  function __construct() {`

MOPS-DanCole:
No problem John...we've got another update coming so I'll be sure to adjust it accordingly....I'll also be interested to hear what Steve discovered...Jim tested it on 7.0 something or another without any issues.

skleiner:
@BrockleyJohn @MOPS-DanCole 
My mistake on that John.  I forgot to enable the development ini file in uniserver...doing that yes there are errors.  Specific to this commit: 
Deprecated: Methods with the same name as their class will not be constructors in a future version of PHP; shoppingCart has a deprecated constructor in C:\Users\Public\UniServerZ\www\Responsive-osCommerce-master\includes\classes\shopping_cart.php on line 13
`  class shoppingCart {
    var $contents, $total, $weight, $cartID, $content_type;

    function shoppingCart() {
      $this->reset();
    }`
If we're dealing with php7 elsewhere, perhaps this would be a good time to take care of the shopping cart error?

BrockleyJohn:
@skleiner my own view would be to keep the two things separate - the testing requirements are somewhat different and it's not really what @MOPS-DanCole signed up for

If (heaven forfend) we find there's something we hadn't thought of in the php7 stuff and it needs a change of strategy, that's going to be a lot harder if it's tangled up with other changes.

Writing and testing new code to (a subset of) php5+ syntax and testing it on php 5.3+ isn't quite the same as changing existing code and testing it on php7.

skleiner:
OK on that.  This level of coding is beyond what I get into normally, so I defer.

raiwa:
@MOPS-DanCole I created a separate pull request for the paypal express error fix

MOPS-DanCole:
Atta boy.

gburton:
Using const is fine.  It is not an error.  Keep discussion to the matter at hand, please.

Looks like this is now an addon:
http://forums.oscommerce.com/topic/409351-modular-shopping-cart/

So this pull is to be closed?

MOPS-DanCole:
No, no....we'll want you to pull it when we're sure it is solid....just wanted to get it out there for people to use and test as quickly as we could. We still want and hope to see it in the core so there will no need for the add-on at some point.

gburton:
Closed.  Thanks to all for putting in a shift on this.

PR #342, opened 2016-04-16
undefined filename error in paypal express

raiwa:
Problem:
When paypal express checkout module is installed and configured,
in [catalog]/shopping_cart.php and [catalog]/checkout_payment.php the following error shows:
Use of undefined constant FILENAME_MODULES - assumed 'FILENAME_MODULES' in ***********includes\modules\payment\paypal_express.php on line 699
Conditions:
php 5.6 or  php 7.0 with error reporting  in application_top.php set to "error_reporting(E_ALL);"

Error may show also on checkout_confirmation.php if returned from PayPal
to confirm order.

Reason:
constant FILENAME_MODULES is defined only in Admin, not in Catalog filenames.php

Suggested fix: 
hardcode filename instead of filename constant.
in: [catalog]/includes/modules/payment/paypal_express.php line 699
Change: FILENAME_MODULES to 'modules.php'

### DIFF
--- a/includes/modules/payment/paypal_express.php
+++ b/includes/modules/payment/paypal_express.php
@@ -696,7 +696,7 @@
       $dialog_error = MODULE_PAYMENT_PAYPAL_EXPRESS_DIALOG_CONNECTION_ERROR;
       $dialog_connection_time = MODULE_PAYMENT_PAYPAL_EXPRESS_DIALOG_CONNECTION_TIME;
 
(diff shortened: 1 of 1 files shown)


skleiner:
@gburton If you're going to consider this for the responsive repo I'll test on uniserver. 
Even though my live shop runs php 5.5 whatever happens regardless I'll likely incorporate this anywway for when my host eventually upgrades php.

gburton:
Thank You

PR #347, opened 2016-04-19
Navbar fixed

gburton:
Optimises the code given by @Piernas - thank you


gburton:
Thank you @Piernas

gburton:
Anyone updating an existing shop will need to make the code changes as shown above, then turn off the module (in admin), and turn it back on (in admin).

PR #348, opened 2016-04-20
Locale example comments update

raiwa:
Comments with locale Examples are outdated ISO_8859-1 references.
They need to be updated to utf8 array examples in the main language files.
Not sure if "ena_au" for Australia is correct, couldn't find reference.


gburton:
@raiwa thanks
Could you possibly clean up your fork and re-submit, as you have lots of unrelated files added here.

raiwa:
@gburton Sorry, have no idea how to do this. It shows for me only the changes in the 2 main language files which are intended. Do you refer to all the unnecessary commits in the list above.

gburton:
Yes, see https://github.com/gburton/Responsive-osCommerce/pull/348/commits

gburton:
@raiwa could you refer me to some reading on this ?

raiwa:
Generic info:
https://en.wikipedia.org/wiki/Locale
https://en.wikipedia.org/wiki/Language_code
http://www.lingoes.net/en/translator/langcode.htm

Some resources for the first 2 definitions:
https://wiki.archlinux.org/index.php/locale
http://www.cyberciti.biz/faq/how-to-set-locales-i18n-on-a-linux-unix/

For the third entrance I found this resource:
https://msdn.microsoft.com/en-us/library/39cwe7zf%28v=vs.100%29.aspx

PR #352, opened 2016-04-25
The term "install"  in store directory name breaks installation process

raiwa:
Problem:
If the store directory name includes "install", the same like the install directory, the wrong part of the path gets stripped for $www_location.
In consequence this produces wrong defined DIR_WS_HTTP_CATALOG and DIR_WS_HTTPS_CATALOG constants in configure.php (store and admin)

Reason:
The function:
  $www_location = substr($www_location, 0, strpos($www_location, 'install'));
searches for the first appearance of "install"

I tested with these directory names:
install_test/
test_install/
install_test_install/
installtest/
testinstall/
testinstalltest/
installtestinstall/
install/
 and got a wrong store path for all of them.

Suggestion to fix the error:
$www_location = substr($www_location, 0, strrpos($www_location, 'install/install.php'));
This function is more secure, it searches for the last occurence of the install directory plus filename.
With the suggested fix all above examples produce a correct result for the path.

PS: Sorry, I still have not discovered how to clean the unneeded commits.


gburton:
Please will you make sure your master is up to date and exact - then branch it off and make code changes.  Please get rid of the change in order.php, and resubmit.  Thanks

PR #353, opened 2016-04-25
Product listing typo

crazy-colorado:
Found a missing " mark in product listing.


BrockleyJohn:
That's not the only difference in this pull request

crazy-colorado:
I found two other issues that I corrected with the same pull request.  With each fix I made comments.  I can recreate the pull; however, the changes are all within this current pull.

gburton:
L166:  you have added itemscope, this is not needed per Schema example markup.
Please resubmit only the missing apostrophe, thanks!

crazy-colorado:
I do agree that my initial research showed that the Schema does not state it is required; however, after some poking around (https://schema.org/docs/gs.html) itemscope and itemtype are used hand in hand.  W3c validation of a product listing page outputted many HTML errors.  Adding the itemscope was the solution.

gburton:
Get rid of the extra markup please.  Resubmit only the typo, thanks.

PR #358, opened 2016-04-27
Jquery version update for install routine

Piernas:
It was updated to 2.2.3 on catalog but not in the install files.


gburton:
https://github.com/gburton/Responsive-osCommerce/pull/356

Issue #288, opened 2016-03-14
PHP 7


gburton:
**PHP7 updates**

Methods with the same name as their class will not be constructors in a future version of PHP in module files and class files

**Admin Side**
- [x] classes - assigned to @BrockleyJohn 
- [ ] functions
- [ ] modules
- [ ] other

**Shop Side**
- [ ] classes
- [ ] functions
- [ ] modules/action_recorder
- [ ] modules/boxes
- [ ] modules/content
- [ ] modules/header_tags
- [ ] modules/navbar_modules
- [ ] modules/order_total
- [ ] modules/payment
- [ ] modules/shipping
- [ ] modules/social_bookmarks
- [ ] other

Piernas:
Do you want them to be done all at once?

gburton:
@Piernas no.  can be done bit by bit (and bit by bit is easier for testing, pushing, and committing).

Piernas:
Ok, I'll do class constructors as I found them.

gburton:
That would be difficult.  Best would be to put your name against one lot of files (eg) /includes/modules/header_tags/ so that others know you are doing those.  And someone else does not then also do it.  Does it make sense ?

Piernas:
Perfect. but I'm seeing it's not much work. I can do all the modules in an hour.

gburton:
@piernas **Please ** do only one lot of modules as I suggested.  Then test them.  Then push...

BrockleyJohn:
ok - put me down for admin/classes
...testing might take a while!

Another php7 pitfall is that functions called statically must  be so declared but I think the only bit of core with any static anything is Braintree payments

If we're doing piecemeal, we'll not be able to test on php7 until a coherent chunk is done. I assume we're just renaming the constructor to __construct (and leaving the documenting to github) and testing that the constructor still gets executed and performs as expected. Will keep a note of what I did for each test.

gburton:
@BrockleyJohn exactly that, and thank you.  admin/classes are yours.

BrockleyJohn:
Actually this approach isn't sufficient; it doesn't cater for the constructor method being called directly as I have found in testing. A general approach to keep changes and testing to a minimum is to create a constructor __construct that calls the existing class name function.

BrockleyJohn:
further - and probably irrelevant - thought:

__construct calling the class name constructor for all classes would preserve backward compatibility to php4
!!

BrockleyJohn:
[...]

PR #361, opened 2016-05-03
Modular index

gburton:
Ready for primetime?  This is only the front-page of index.php

I'd like to get this into Core (even though index.php is now a mix of modular and "old skool", as it'll take some time to update the other pages (categories, manufacturers).   Anyone for or against?  If against, for what reason(s)...


skleiner:
I'm good with proceeding. Perhaps a wiki page on this aspect of the project, noting the level affected, so there's no confusion (like I was) to the user.  When the other levels are integrated then the wiki updated?

azpro-git:
> Ready for primetime? This is only the front-page of index.php
> 
> I'd like to get this into Core (even though index.php is now a mix of modular and "old skool", as it'll take some time to update the other pages (categories, manufacturers). Anyone for or against? If against, for what reason(s)...


Definitely a go!

flyerart2B:
Hey there... have we given up on resolving php7 problems? The new modularity is great but oscommerce responsive still does not work on php7 which would seem a priority. Sorry, I had a serious car accident and have been out of the loop so maybe I missed something.

gburton:
@flyerart2B https://github.com/gburton/Responsive-osCommerce/issues/288
Join in the discussion there.

azpro-git:
@gburton Looking at Installer it reminds me of a (critical / potential ?) issue for the setup/installing of modules.
With the modularity we heavily rely on configuration setting MODULE_CONTENT_INSTALLED. If not all settings / permissions are set correctly by shopowner or Installer (possible with different Server settings) the reading of Installed modules can be faulty. Now on line 87 in admin/modules_content.php it reads:
```
if ( implode(';', $_installed) != MODULE_CONTENT_INSTALLED ) {
    tep_db_query("update " . TABLE_CONFIGURATION . " set configuration_value = '" . implode(';', $_installed) . "' where configuration_key = 'MODULE_CONTENT_INSTALLED'");
  }
```
So if $_installed is empty ... MODULE_CONTENT_INSTALLED will be set to empty .. which leaves us with an not working store.

I highly recommend to add   if ( !empty($_installed) ) {  before Line 87.

This off course is also applicable for modules.php

I did some months ago a lot of testing for this issue and cannot remember exactly in which cases $_installed can be empty - But I did learn (the hard way) it can happen!
I also added some $messageStack->add() messages to know / see what is happening in LINE 88 (is MODULE_CONTENT_INSTALLED updated or not) but that is probably just me being overcautious.

As you probably know - I am sorry I don't know how to make a Pull-request. Learning Github is now to much while setting up the new BS store. Still hop this helps!

gburton:
Thanks to all who took the time the test and to post relevant comments.

Issue #354, opened 2016-04-26
Address book fails


dbusr:
To see the problem, enable the gender option for the customer and create a second address book entry. Then checkout and on the checkout shipping page click on the Change Address button, select one of the existing entries and click continue. It won't let you return because it fails on the gender entry for a new address.  It's doing that because the modules/address_book_details.php has teh following for gender: required aria-required="true"

gburton:
Thanks for reporting this.  Please use code changes at https://github.com/gburton/Responsive-osCommerce/pull/355 and report back.

dbusr:
Hello,

Yes, removing the required parameters allow it to work.

g> Thanks for reporting this. Please use code changes at #355 and report back.
g> —
g> You are receiving this because you authored the thread.
g> Reply to this email directly or view it on GitHub 



                      


---
This email has been checked for viruses by Avast antivirus software.
https://www.avast.com/antivirus

PR #372, opened 2016-05-11
 Checkout Address Book -> Remove required files.

katapofatico:
The problem reported on #354 it's fixed when we apply the code changes at #355 on checkout_new_address.php. This commit do it.


katapofatico:
Excuse me my redundant commits :)

gburton:
Thank you @katapofatico

PR #373, opened 2016-05-12
Modular index nested aka Categories

gburton:
Modularises another part of index.php


gburton:
Any and all RELEVANT comments (and code changes if appropriate) appreciated.  
Thanks

BrockleyJohn:
Installer needs the modules adding to MODULE_CONTENT_INSTALLED too
Currently having some other issues but I think they're probably me - will come back to it fresh

BrockleyJohn:
The other issues I was having were down to testing against a 'products' depth category instead of a 'nested' one. Perhaps once the same modules exist for 'products' depth too, and if they're all enabled by default, the potential for confusion will be reduced... see how less tired people get on.

I wonder if now might be the time to revisit the definitions of categories in this context?
currently:
containing only categories =  nested
containing only products = products
containing products and categories = products
containing neither products nor categories = products

gburton:
There is 3 parts to index.php;

1. top
first page you see in the site.
2. nested
categories and subcategories, with some products shown
3. products
shows products from a category that has them or from a manufacturer

In other words, to get to a list of products, you navigate through the nested area.

Let's not think too hard about this ;)

skleiner:
@gburton Did an install on uniserver...prelimary testing installed OK.  Testing nested modules ran into issue with Category Description.  Still shows after disabling and uninstalling.
Am I looking at the correct thing here? In admin->catalog->categories/products when editing a category description "Edit the Category Description:" if I enter text in the box below? Is that what the index_nested module is controlling?

gburton:
@skleiner Steve thanks for testing.

http://template.me.uk/234bs3/index.php?cPath=1 is nested as it has subcategories.
http://template.me.uk/234bs3/index.php?cPath=1_4 is products as it has no further subcategories

Nested is modular, so if you turn off the description module it won't show.
However on a products part of the page, it would always show as that is not (yet) modular.

So, is the page you are looking at "nested" or "products" ?

skleiner:
OK...now I understand.  index is a complicated matter.
So yes, in this case I was at the product level.  Hardware->Graphics Cards

So i understand this first level category (non nested) description - no products will not show.
Second level category (nested) description, showing sub categories - no products the product description should show if entered in admin.?
Third level category description showing only products will show always..

gburton:
Pretty much the rule is this:

Does the category you are looking at have further subcategories;
YES:  it should contain no products
NO:  it should contain products...

I have seen sites with 7 levels of nested!

BrockleyJohn:
Here's a prediction for you: this is going to be _at least_ as annoying with false bug reports as 'equal heights'. It's much less than obvious to a new shop admin - and apparently not so new ones ;)

If we carry on with nested & products, I know I'm going to be writing a module for 'products' categories that allows navigation to categories below... or at least highlights their existence with a message. I've seen too many "broken" stores that just had a few stray products in the wrong places, hiding the lower levels in the explorer view. I just thought the concept could be extended to avoid duplicating modules between nested & products, and to make more sense to a shop owner.

Anyway - I've now said my piece, am aware of the danger of overthinking this and am ready to do my part in fielding the problem reports on the forums.

skleiner:
From a 95% shopowner / 5% code tweaker point of view, the index thing is different for sure. Whereas mostly the content modules are if you enable this, then that happens, but with index, if you enable this, then that happens, unless condition B exists, then something else will happen.  Once explained to me, I'll get it, but it's not intuitive especially if you haven't mastered what the pertinent code your looking at is actually doing.

That being said...onward with testing....and maybe related to what I just said....

I added two sub cats to the test shop, I now have:

Catalog> Hardware >CDROM Drives showing

sub cats of 
New CD Drives
Used CD Drives

I set up a category description for Edit the Category Description for the New CD Drive category:  it says:  "Testing New CD Drive Text" on the edit page.

Back in test shop, on this page;

http://localhost/Responsive-osCommerce-Modular-Index-Nested/index.php?cPath=1_17

Which is CDROM Drives there shows the two sub cats I just made.  But there is no description for the New CD Drives sub.  Beside it shows Used CD Drives with no description, but there is none anyway.

I have the index_nested Category Description module enabled

If I drill down to New CD Drives. I see the description which as per previous discussion is expected.

gburton:
Take this Example:
Catalog [top] 
> Hardware [nested == has subcat] 
>> CDROM Drives [nested == has subcat]
>>>New CD Drives [products = no subcats]
>>>Used CD Drives [products = no subcats]

Therefore...

if on hardware, page is modular
if on CD ROM Drives, page is modular
If on New CD Drives, page is NOT modular
If on Used CD Drives, page is NOT modular

ASS-U-ME shopowner has inserted a Category Description on all these categories...
On the modular pages (aka NESTED) it will show only if the relevant module is installed
On the non-modular pages (aka PRODUCTS), it will always show until it is modularised at which point the shopowner will need to install the relevant module...

I don't see the complication...

skleiner:
OK. I have it now. I was looking for the cat description to show below/with each subcat in the grid.

Now that I see this works for just the category you are on.  Not for those sub cats displayed below.

So all's right with the world again.

skleiner:
Further testing, new products can be manipulated by max number displayed and product width only in nested mode, i.e. when on a category and only sub cats are displayed.  Good things.

However on catalog level, the New Products For (Month) reverts to another module and is controlled by setting elsewhere? Meaning at catalog level, I have 6 new products displayed.  At sub cat level I have 3 displaying (as per module setting)

In my **standard responsive test shop** I see that catalog level new products display can be set in configuration, max values:  Maximum number of new products to display in a category. 

I don't see that setting in my test shop for nested.

BrockleyJohn:
@skleiner the setting for the top level is in the content / index / new products module settings

skleiner:
@BrockleyJohn Thank you. And so it is.

skleiner:
And finally....Category Listing, index_nested works as expected. Again, being able to chose category width is a plus.

skleiner:
So, all that being said, the technical stuff John was addressing I don't understand, but aside from my personal learning curve dealing with what I've been used to in index vs how to use index modular, I favor continuing modularization.  Going back into standby mode....

gburton:
> I just thought the concept could be extended to avoid duplicating modules between nested & products, and to make more sense to a shop owner.

Sure thing.  I believe it makes sense to have separate (although extremely similar) modules as seen in "index -> new products" and "index_nested -> new products".

Reasoning;

1.  It makes things easier to code
2.  It makes things easier for the shopowner (example; show 6 new products on index, show 3 new products on index_nested)
3.  It makes things easier for amateur coders to over-ride.  No fear of mucking up "top" when tryinh to do some new module for nested

and so on.  The problem may come as there are not many who realise that index is split out into 3 sub-indexes, and one of those subindexes deals with two things;

top
nested (always categories)
products (could be products in a category, or products of a manufacturer)

Anyway, thanks chaps for testing and comments :+1: 
Will drag this into Core after the weekend.

gburton:
Thanks @BrockleyJohn @skleiner for testing and commenting

PR #376, opened 2016-05-17
Minify CSS

gburton:
Helps with PageSpeed (slightly)

There is nothing added or removed, just a minification of the CSS


gburton:
Thanks for testing @skleiner 
Adds some CSS for display (that is not linked to Bootstrap)

PR #377, opened 2016-05-17
Navbar Text

gburton:
Per http://getbootstrap.com/components/#navbar-text
navbar-text class should not be p not li


gburton:
Thanks for testing @skleiner

PR #391, opened 2016-06-12
Tap Targets

gburton:
Bigger Tap Targets make life easier for those on mobile phones.


skleiner:
The bigger tap targets do help on the smartphone,  Nice to have the nav-pills features. In my case I have five information links stacked and on desktop the spacing pushed the box down pretty far. I tweaked the spacing back a tad with some padding adjustments.

PR #392, opened 2016-06-13
Historical clean up

gburton:
These commits do away with a bunch of historical html and .css in favour of Bootstraps own .css

This makes the HTML slightly smaller, and the custom.css a lot smaller.  Which is obviously good for page speed / loading.


skleiner:
All CSS display updates appear to be working fine after install onto live shop.

PR #415, opened 2016-08-11
Global Vars

gburton:
$HTTP_X_VARS to $_X


baulie38:
Changes, why not also in the admin area?

gburton:
On the to-do list @baulie38 but that list is long and time is short - all done in spare time.

PR #419, opened 2016-08-16
remove include filenames.php from admin application_top

BrockleyJohn:
Include still seems to be there (but the file isn't)...

...or did you mean to leave an empty file for the sake of old addons?

Either way, it doesn't work as it stands.


gburton:
Thank you @BrockleyJohn

PR #420, opened 2016-08-16
Hardcode Paths

gburton:
Gets rid of (most of) the DIR_WS_* in favour of hardcoding.


BecRoM:
catalog/install/templates/pages/install_4.php contains:
`'  define(\'DIR_FS_DOWNLOAD_PUBLIC\', DIR_FS_CATALOG . \'pub/\');' . "\n\n" .`
and was harcoded

gburton:
That is correct use.  DIR_FS != DIR_WS

PR #422, opened 2016-08-19
editing errors in filename constant deprecation

BrockleyJohn:
string replacements done in the wrong sequence in admin/orders - corrected


gburton:
Thank you @BrockleyJohn

PR #427, opened 2016-08-25
separate multiple fields output from selected payment module

BrockleyJohn:
No biggie but on checkout confirmation, multiple fields output from selected payment module were coming out:
Label1 Value1Label2 Value2Label3 Value3...

Now:
Label1 Value1
Label2 Value2
Label3 Value3
...


gburton:
Thank You

PR #430, opened 2016-09-03
Php7 consolidated

BrockleyJohn:
All the changes consolidated into one with reported and unreported merge conflicts resolved (there were plenty); for ease of further testing and implementation.

Outstanding issues:

presumably obsolete & not included
admin/includes/classes/payment_module_info.php paymentModuleInfo

unused in core but still included (may break addons if removed?)
catalog/includes/classes/boxes.php tableBox, infoBox, infoBoxHeading, contentBox, contentBoxHeading, errorBox, productListingBox

reportedly unfit for purpose (and untested)
catalog/includes/classes/http_client.php httpClient
only used in the equally unfit and untested usps shipping method

moneybookers payment module can't be tested without a valid account, which they refused to give me. Now known as skrill. Does module still work?


gburton:
I cannot use this due to Raiwa's commits.
Sorry.

BrockleyJohn:
Would you prefer the changes consolidated & ready to whack in? I can redo Raiwa's branch & remerge

PR #443, opened 2016-09-27
Hooks Implementation

gburton:


gburton:
Discussion welcome.  Testing welcome.

gburton:
Next up after this will be adding of the Paypal App.

skleiner:
My includes/application_top.php is not up to date with the most recent Edge but I don't think it matters because I already had the code for hooks integrated there from some earlier work.  So I just updated all the "hooks" files with the commit files. The existing hooks I have are functioning normally.  On the admin side, the hooks box is displaying correctly.

BrockleyJohn:
Tested on clean install. Empty hooks admin page shows as expected. Catalog side works fine (without any hook modules!)

What about admin hooks?

gburton:
> Hooks class constructor is php4 - update to __construct

Done, please test - thanks



> What about admin hooks?

Do any exist?

BrockleyJohn:
Paypal app uses one on admin/orders

gburton:
> Paypal app uses one on admin/orders

Paypal App is not yet core...

BrockleyJohn:
neither are any hooks :) - I just thought if you're going to put it into app top on the shop side, and if paypal app is next up, it might be clearer to put both includes in at the same time.

Unless I'm missing something, there aren't any listeners anyway...

gburton:
All Hooks load from the shop side...

BrockleyJohn:
I guess you mean more than that they live under cat/includes not admin... I'll have to wait to see one

gburton:
> I guess you mean more than that they live under cat/includes not admin... I'll have to wait to see one

All hooks live in the shop side (even the admin side hooks).  Have a look at the Paypal App Hooks.

BrockleyJohn:
> All hooks live in the shop side (even the admin side hooks).
I guessed wrong then! 
> Have a look at the Paypal App Hooks.
I have been - but have only seen the one on admin/orders. Couldn't find any shop-side examples. Am I looking in the wrong places?

gburton:
paypal_app (from addons) > unzip.
look in /includes/hooks/

BrockleyJohn:
contents is
/admin/orders/paypal.php

gburton:
Yes...
Which suggests what ;)

BrockleyJohn:
That we're not participating in the same conversation? 

Surely that's the hook on admin/orders not a shop-side example...

gburton:
/includes/hooks/admin/orders/paypal.php

That is an ADMIN hook, which lives in the include/hooks/ directory SHOPSIDE.

BrockleyJohn:
Yes, I knew that before the first time you said it :) It's the only hook I've ever seen.

I meant an example of a hook that works on the shop side!

gburton:
LOL,ok,  There are examples at the osc forum.

PR #444, opened 2016-09-27
Equal Height

gburton:
New approach as per http://forums.oscommerce.com/topic/408314-equal-height-possible-new-approach/

If you want to test this on your live shop, ensure that you FIRST uninstall the equal heights header_tag module.  You then only need to add in the .css changes (better to do this in user.css until you are happy with your testing, then move to custom.css after).  You do not actually need to delete the HT module.

If you do not like the new approach, simply re-install the HT module and remove the .css you added.


Piernas:
Not working here nor in demo page http://template.me.u....php?cPath=3_10 was this one updated?

gburton:
Yes.  You probably need to test in a number of browsers and devices.

skleiner:
With the changes made in user.css and the equal heights module uninstalled,  all seemed to be working fine with the exception of a featured products add on I have.  Adjusting the class in the template file for that gives something acceptable.

gburton:
@skleiner thank you for testing and commenting

skleiner:
Ran into an issue with the new products nested module.  The products show a full row of three,  then blank, blank, product.  The next row shows product, product, blank.  This is with 6 products set to show and a width of 4.  Reinstalling the equal height div module in HT corrects this.  I note there is a better display overall with the ht module vs the css.

gburton:
@skleiner can you recreate that on this page: http://template.me.uk/234bs3/index.php?cPath=3_10
and take a screenshot?  Also what device and what resolution (screen size)...  Thanks!

gburton:
Or maybe you mean here:  http://template.me.uk/234bs3/index.php?cPath=3
If that is the case, this particular module doesn't utilise the new .css settings (yet)

skleiner:
That class code did fix the issue with the nested new products.  However,  the outline of the box for each product is uneven in my shop with just the css code used.  Meaning the boxes grow to different heights depending on the content and they don't line up evenly side by side. With the ht equal heights module activated all boxes are the same height and line up evenly.

gburton:
`.productHolder {height: 100%;}`

skleiner:
Testing...this helps but not everywhere.  Not working on index page...new products module.  However does work on nested category new products module when I modify line 23 to this:
`      <div class="thumbnail equal-height productHolder">`
If I test this in the featured product add on module, I get partial success by adding the class call in there.

gburton:
Thanks for testing Steve.

BrockleyJohn:
Sometimes the small things are the biggest. This is my favourite change for ages - roll on bootstrap 4 and we can bottom align the products so that the buttons line up.

gburton:
@BrockleyJohn
:+1: 

@skleiner
The display problem is preferable to the .js problems in my opinion.  Of course can still use the .js solution should shopowner wish to.

gburton:
Forgot to mention that we also need to find other lists of products and place them into a list-group.
Those commits are upcoming as and when shopowners report them @ https://github.com/gburton/Responsive-osCommerce/issues/446

PR #447, opened 2016-10-11
Paypal App

gburton:
http://addons.oscommerce.com/info/9184

Updated for Edge.


gburton:
Of particular interest for testing is the new admin/orders.php page...

skleiner:
I have this installed on uniserver and will try testing there

Piernas:
Seems to be working here. Only issue I see: The tabs have hard coded titles, and should be inserted into language files.
Another good point would be to have stock tabs into separate hooks so the stock ones can be replaced easily and there would be no code changes in orders.php to modify files.
Also an id on the infobox buttons would allow to add a new button from a hook.

Piernas:
Also found there are some references for 2.2 compatibility still in the code.

skleiner:
From uniserver,  having difficulties with retrieving credentials, both live and sandbox...after going through the process get to a page that has a link to "return to osCommerce"..click that and go a site starting with https://ssl.oscommerce.com/index.php.....and all the page shows is rpcStatus=-100
Maybe has something to do with uniserver..don't know.
Manual credential entry appears to work.

skleiner:
Made successful sandbox purchase, PP express.  On customer side..order not showing in account_history, "You have not yet made any purchases"
On admin side, admin/orders.php order shows.
From the admin orders page, using the PayPal tab tried an refund in the sandbox purchase...appeared to work OK

BrockleyJohn:
From the original app, not this version: I don't think the retrieve credentials works in UK, actually. I think they changed the Paypal end of things and it doesn't feed through properly. It needs a change on the oscommerce.com end of the app to fit with that. I gave up on the retrieve in the end.

Piernas:
I can retrieve sandbox credentials, but I get same error (rpcStatus=-100) on live credentials - with a spanish account.

Kymation:
I am unable to retrieve credentials on a live server. I always get rpcStatus=-100. This has never worked for me using the previous version either, so it's nothing new.

BrockleyJohn:
There's quite a lot of php4 > 7 stuff to do on this; to the extent that I think it would better be left until after it's in master and handled along with the older stuff so it can be properly & cleanly tested. What do you think @gburton ?

BrockleyJohn:
please see pull [https://github.com/gburton/Responsive-osCommerce/pull/448](https://github.com/gburton/Responsive-osCommerce/pull/448) for fixes for known problems
(ref [http://forums.oscommerce.com/topic/408938-paypal-app-api-credentials-and-retrieve-v-manual-setup/](http://forums.oscommerce.com/topic/408938-paypal-app-api-credentials-and-retrieve-v-manual-setup/)

gburton:
Thank you all for testing this so thoroughly - appreciated.

gburton:
TY to all

PR #448, opened 2016-10-12
tab texts to language file

BrockleyJohn:
remove hard-coding of non-paypal tabs on admin / orders


BrockleyJohn:
added fixes for three known problems with PP app:
- undefined constant error on initial setup (Start)
- UK v GB mismatch between app and osc country table
- force curl to TLSv1.2 instead of SSL (the default on some versions)

gburton:
TY

BrockleyJohn:
done

gburton:
TY

gburton:
Could you send your changes to paypalApp branch please.
You've managed to send to a very old branch of similar name.  My fault for not deleting the branch.
TY

PR #449, opened 2016-10-16
Paypal app bugfix again

BrockleyJohn:
Tidying up, I found out the undefined constant warning in ppApp is not restricted to Start processing.

If you install a module then uninstall it, you get the warning on accessing the Paypal page in admin.

If you uninstall a second module it leads to a headers already sent error and you can't access the page.

PR for that fix, the country table and the tab titles to language file.


gburton:
TY John.
Sorry for my earlier mistake on the incorrectly named Branch.

Issue #451, opened 2016-10-23
Structured data  in module also_purchased_products.php


azpro-git:
http://template.me.uk/234bs3/product_info.php?products_id=28 for also purchased Google test tool:

https://search.google.com/structured-data/testing-tool#url=http%3A%2F%2Ftemplate.me.uk%2F234bs3%2Fproduct_info.php%3Fproducts_id%3D28

Gives an error:
ItemListElement
	
SWAT 3: Close Quarters Battle (Must have Value  in field position.)

gburton:
http://pastebin.com/Y5cNHqEK

Please test, thanks

azpro-git:
Tested! Looks fine now!

gburton:
https://github.com/gburton/Responsive-osCommerce/pull/453

PR #452, opened 2016-10-24
Update also_purchased_products.php

gburton:
Corrected Schema


gburton:
Further reading:
https://github.com/gburton/Responsive-osCommerce/issues/451

PR #459, opened 2016-11-21
allow uppercase on SQL sintax

katapofatico:
http://stackoverflow.com/questions/608196/why-should-i-capitalize-my-sql-keywords

The simplest reason: why not?

### DIFF
--- a/includes/classes/split_page_results.php
+++ b/includes/classes/split_page_results.php
@@ -32,18 +32,18 @@
       $this->number_of_rows_per_page = $max_rows;
 
       $pos_to = strlen($this->sql_query);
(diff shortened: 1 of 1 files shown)


BrockleyJohn:
Someone sticking up for us old COBOL programmers? Whatever next? Hierarchical databases?

katapofatico:
I'm not old COBOL programmer. It's a question of readibility, not fashion, school or origin.
Perhaps this topic has already been discussed. If it's so I',m sorry much.

gburton:
TY.

You need to get this into the v2.3.5 branch of the main osCommerce build.
Once that is achieved, it would then be ported over to this build.

PR #468, opened 2016-12-15
Missing closing ?> tag

rudolfl:
Some housekeeping.
File is missing closing php tag.


gburton:
PHP Files (that do not output HTML) should not have closing ?> tags

gburton:
Uhhhhh

PR #471, opened 2016-12-30
$cPath_array must be initializated

katapofatico:
Several files call like this:
`(sizeof($cPath_array) > 0) `
Example: [admin/categories.php](https://github.com/gburton/Responsive-osCommerce/blob/master/admin/categories.php#L976)
...therefore $cPath_array must be initializated

### DIFF
--- a/admin/includes/application_top.php
+++ b/admin/includes/application_top.php
@@ -216,6 +216,8 @@
   } else {
     $cPath = '';
   }
+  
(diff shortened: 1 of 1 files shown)


gburton:
TY.  You need to get this into the 2.3.5 osCommerce.  No code changes are supported at this time here, unless they are specific to the changes we made adding/improving the Bootstrap.

PR #476, opened 2017-01-10
Undefined session id fix

raiwa:
admin/includes/functions/html_output.php throws undefined variable $SID
in line 48.
Shows only if   error_reporting(E_ALL);
in admin/includes/application_top.php line 17


gburton:
TY. You need to get this into the 2.3.5 osCommerce. No code changes are supported at this time here, unless they are specific to the changes we made adding/improving the Bootstrap.

Issue #477, opened 2017-01-11
(int) 0 is not null


katapofatico:
In some situations that I don't understand exactly, the checkout_shipping.php choose the cheapest shipping method:

```php
// if no shipping method has been selected, automatically select the cheapest method.
// if the modules status was changed when none were available, to save on implementing
// a javascript force-selection method, also automatically select the cheapest shipping
// method if more than one module is now enabled
  if ( !tep_session_is_registered('shipping') || ( tep_session_is_registered('shipping') && ($shipping == false) && (tep_count_shipping_modules() > 1) ) ) $shipping = $shipping_modules->cheapest();
```

cheapest() method checks this before comparing and collecting the costs:

```php
if (isset($quotes['methods'][$i]['cost']) && tep_not_null($quotes['methods'][$i]['cost'])) {
```

If the shipping cost is 0, then tep_not_null() returns false, skipping the shipping method involved and causing several problems.

I have solved this problem in my shop this way:

```php
if (isset($quotes['methods'][$i]['cost']) && (is_numeric($quotes['methods'][$i]['cost']) || tep_not_null($quotes['methods'][$i]['cost']) )) {
```

What's your evaluation about this problem? Should I save 0 shipping cost any other way in order to avoid it? Is it a bug? Is my solution the best solution?

Probably related: [tep_not_null on admin](https://github.com/gburton/Responsive-osCommerce/commit/dbe473677f03f42bd605cbb5da5ec440cd6440af).

gburton:
0 is not null, so it's correct...
Your solution looks good to me.

katapofatico:
@gburton Thanks for your response, I go to push the fix.

gburton:
TY.  You need to get this into **official** 2.3.5

There will be no further code changes here unless the code changes are to boost what we have done in the Responsive changes.

PR #478, opened 2017-01-16
(int) 0 is not null #477

katapofatico:


gburton:
TY.  You need to get this into **official** 2.3.5

There will be no further code changes here unless the code changes are to boost what we have done in the Responsive changes.

PR #481, opened 2017-01-19
Responsive os commerce spanish language files

raiwa:
Should be all up to date.
All translated except payment modules.
Only the following payment modules are translated:
- cod.php
- sage_pay_form.php
- sage_pay_server.php

PayPalApp:
- paypal_express.php
- paypal_standard.php
- paypal_login.php


gburton:
TY, I can use these translations in the Espanol branch,

PR #499, opened 2017-02-02
Update credentials.php

Piernas:
Note I don't translate terms as "Live" or "sandbox" as Paypal doesn't do it in most cases.


gburton:
TY

PR #514, opened 2017-02-07
Update verify_ssl.php

Piernas:
@gburton I'm not sure if I must translate 'True' and 'False' - I suppose these values are directly used in compares, right?


gburton:
TY

PR #529, opened 2017-02-13
Remove Box Class file

gburton:
Boxes are no longer created using this file in the SHOP side.


nukuaes:
Any problem, work right.

mcmannehan:
i can agree with you. the box classfile not used anymore

PR #534, opened 2017-02-27
Bugfix: WorldPay Hosted Payments - check query overwriting orders class - rename check object

BrockleyJohn:
Bug.

Symptoms: empty list of products in order email and stock not getting updated

Cause: check query on order status in before process was overwriting the orders class, with knock-ons through the rest of it.

Fix: rename result of query to something other than $order

### DIFF
--- a/includes/modules/payment/rbsworldpay_hosted.php
+++ b/includes/modules/payment/rbsworldpay_hosted.php
@@ -349,11 +349,11 @@
         tep_redirect(tep_href_link('shopping_cart.php'));
       }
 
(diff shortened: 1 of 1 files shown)


BrockleyJohn:
@gburton please see [https://forums.oscommerce.com/topic/411165-missing-products-and-not-updating-quantity/](https://forums.oscommerce.com/topic/411165-missing-products-and-not-updating-quantity/)

gburton:
Needs to be fixed at main repo first...

PR #537, opened 2017-03-09
Linuxuk patch 1

linuxuk:


gburton:
Please push to BrockleyJohn:php_compatibility_02
TY

BrockleyJohn:
This is actually a MySQL 5.7 issue rather than a php7 issue - not reproducible on php7 + MySQL 5.6.21
- basically you now need to include any colummns in ORDER or GROUP clauses in the SELECT

Perhaps we need a MySQL 5.7 compatibility issue to collect problem reports?

BrockleyJohn:
> Perhaps we need a MySQL 5.7 compatibility issue to collect problem reports?

We don't. I've just been through and tested the **core** code against MySQL5.7 without finding any errors - so even if there are some, I'm confident there aren't many.

Issue #535, opened 2017-02-28
Safari


MyBookShop:
I've just noticed that on Firefox and IE the pages render 4 products aside in Window and Ubuntu (all good).

In Safari and Firefox on my Ipad Air the front page renders like on an android or iPhone all products in one long line down, this also happens if more than one product is in a category.
In portrait mode it renders three products side by side then four. (No equal height mod installed which fixes the  3 then four but installed does not fix the portrait mode)

There is room on the Ipad display as the logo and search bar are side by side as normal.

This did not always happen and may have been broken recently?

I re-installed a clean copy of edge from the repo 2 days ago
and the clean copy also shows like this with no changes.

at a guess it may be bootstrap:-

http://getbootstrap.com/examples/grid/

BrockleyJohn:
I would expect the tablet display in portrait to be the same as phones, but landscape to be the next one up.
From memory, the phone/tablet breakpoint is at 768pixels and the tablet/desktop breakpoint at 992px
You should be able to get exactly the same behaviour on the desktop by reducing the width of the browser window

BrockleyJohn:
[http://whatsmy.browsersize.com/](http://whatsmy.browsersize.com/)

BrockleyJohn:
@MyBookShop  Do you see the same on all pages that show a bunch of products?
- top level index (new products)
- within a category (new products in category)
- index products within a category
- new products page
- specials page

The first two have not been changed for ages, the last ones I changed last month and they may need tweaking. I don't have access to a recent Apple device and they behave as I expected on my ipad 2.

gburton:
I missed this conversation.  What's the outcome @MyBookShop @BrockleyJohn

MyBookShop:
I'm sorry.

I might have a dodgy I pad Air.

I will have access to another i-pad tomorrow.

The test was a clean/fresh install www.143books.com

It displays ok-ish in landscape but in portrait its like an I-phone when
displaying the New Products grid on the front page. The header and footer
etc are correct.

If you want to have a look and let me know.

Ta

MyBookShop:
Yes it still displays badly on a new Ipad Air 2

The product grid is at fault.

It works better in landscape.
It did not do this a few revisions ago.

Ta.
Mark

MyBookShop:
> I would expect the tablet display in portrait to be the same as phones,
> but landscape to be the next one up.
> From memory, the phone/tablet breakpoint is at 768pixels and the
[...]

PR #538, opened 2017-03-14
Fix Safari problems with flex (grid) layout. 

BrockleyJohn:
Fix for issue #535 
Blank column down right hand side on all versions of Safari due to before/after pseudo-element handling. 
Some older versions (notably on iPad Air) not wrapping or not equalising height, so added backwards compatibility (webkit styles).


BrockleyJohn:
Think I had a caching issue in the iPad Air 2 emulation and testing was affected by previous content of user.css
iPad Air 2 and Galaxy Tab S2 seem to require webkit flex styles in a separate rule.

gburton:
Is this a fix for Safari where Safari have not got things correct?
In other words, should Safari be making relevant changes?

BrockleyJohn:
I see where you're coming from. There are actually two problem here: 
- one in which Safari gives an empty column down the right hand side in, I think, all versions. This is due to Safari treating the before and after pseudo elements differently than other browsers do but how they should be treated isn't standardised. This could perhaps be seen as a Safari bug but the "fix" is simply to style the before and after elements to make sure they come out how we want, which might be a good thing if the default behaviour of browsers were to change in the future.
- the other problem where some browser versions don't wrap the elements but stick them in a long column. This was initially reported on Safari but then Carine had the same on Chrome. Putting in the older webkit rules makes the styles work on all browsers that support that earlier version of flex.

gburton:
What's the very latest URL for testing?

BrockleyJohn:
Code is the same on both, data maybe better on 
[http://bromleybr.co.uk/bsdev/](http://bromleybr.co.uk/bsdev/)

gburton:
that breaks quite badly in Win10 Chrome...can't work out how to add an image here, so I popped it onto the BS thread at the forum...

BrockleyJohn:
[http://bromleybr.co.uk/reflex/](http://bromleybr.co.uk/reflex/)

Are you looking for any particular behavioural change?

gburton:
@BrockleyJohn  there should be no behavioural change at all.

A mistake was made by us in the early days where we used list-group for the product listing grid/list thing.  this has caused an incompatibility with http://getbootstrap.com/components/#list-group

Changing out .css to .row.list-group solves it.  Makes sense?

gburton:
TY

PR #540, opened 2017-03-19
Update english.php

LBManiac:
Added placeholders for login screen


gburton:
TY, I am not accepting code changes unless they are applicable to 95% of shopowners.

Language changes should be made on a site by site basis.

nickwuk:
Nice project! Just thinking if you were to re-structure with a template system like BTS then different shops could use different templates making more flexible for such updates.

gburton:
@nickwuk thanks.  BTS is rubbish, always has been, always will be.

frankludriks:
@nickwuk All you need to do is change the CSS in this version of osCommerce and you can come up with something brilliant.

Here are some possibilities just by changing ONE file https://www.tonerpak.com.au/osc/

gburton:
My previous reply was unhelpful, apologies.  

Is there a **real** need for a template system?  Take a look at any shop running the old STS, BTS systems and you will see that their owners made a design, then never changed it.  Ecommerce is not like a blog;  on a blog, branding/style does not matter, on a shop once you have a design in place ... it usually stays the same (or very similar) forever.  Is a Template System needed? 

Did you take a look at the commercially available Responsive Templates - each of these changes no core code, yet look (not just colours, but also layout) radically different to the out of the box installation. 

I've made it easy for anyone to come up with a good design, and (with a small amount of technical knowledge) get that into osCommerce...

Is a Template System needed?  Maybe yes - but state precisely why it is needed...

nickwuk:
I've used BTS for over 10 years and what I liked about it, after further modification, was that it allowed most of the php code to be separated from the layout html, and that as a developer I could use a customised core fileset with different templates for different sites; Also that a site revamp could be demonstrated and tested by switching templates back and forth; and that changes to the outer template code are easily done in one main file (I realise there is template_top and template_bottom these days). 

However I'm not up to speed with the latest developments working with the bootstrap templates .... @frankludriks thank you for the link! Clearly there's already much investment in templates using the existing core structure, however many other systems have chosen to use a template system so they must consider it worthwhile too.

I think what you're doing with this project is excellent, although I'm not sure if it's enough for me yet to return to using osCommerce even in responsive bootstrap form unless it uses a template system, and for page content to be stored in the database and editable via admin ... but that's another topic.

BrockleyJohn:
Have a skim of this thread, particularly the linked post:
[http://forums.oscommerce.com/topic/410171-24-series/?p=1749308](http://forums.oscommerce.com/topic/410171-24-series/?p=1749308)

nickwuk:
@BrockleyJohn many thanks, now starting to understand how development is progressing.

PR #551, opened 2017-05-07
Java-Script Check payment method - maybe transfer to messagestack

rubberfashion:
Hello,

i am not a professional and can also give no possibility for change.
Problem is, that most of the purchases come from mobile. There the checking of possible payment methode is checked by javascript.
When the custome forget to select one or it is not preselected he get an JS error and have to reload the page (You can't continue anymore). This cause many leaves from the website. 

Also other errors are created with JS.

Itsn't it better to change this to existing messagestack?

example:
class/payment.php

 $js .= "\n" . '  if (payment_value == null) {' . "\n" .
               '    error_message = error_message + "' . JS_ERROR_NO_PAYMENT_MODULE_SELECTED . '";' . "\n" .
               '    error = 1;' . "\n" .
               '  }' . "\n\n" .
               '  if (error == 1) {' . "\n" .
               '    alert(error_message);' . "\n" .
               '    return false;' . "\n" .
               '  } else {' . "\n" .
               '    return true;' . "\n" .
               '  }' . "\n" .
               '}' . "\n" .
               '//--></script>' . "\n";


BrockleyJohn:
Are you saying that the payment method selection page doesn't work on mobile?

I will attempt to reproduce.

rubberfashion:
Hello,

it works, but when you did not choose and click continue you will get a JS messagebox. When you click "OK" you have to reload the page to check again. Much better it is to maybe include in messagestack.

BrockleyJohn:
Incidentally, it would be better to have raised an issue, not a pull request.

A PR is for when you have made some changes that you think are worth putting into the core code.

When you have experienced a problem, please raise an issue.

BrockleyJohn:
If you have to reload the page before you can choose a payment method, then it doesn't work properly. The whole point of javascript validation is to get the page right before it goes to the server.

rubberfashion:
OK, Thank you :)

rubberfashion:
The JS validation works perfectly. But when you click ok in the JS error, you can not continue or select the payment method. The continue button is grey and you can not click on it. If you want to proceed you have to reload the page. I solved the issue by preselecting a payment method. So that JS error can not occure

BrockleyJohn:
You should not be getting this behaviour. The core code does not do this (at least not on my Android mobile). Please try it on [http://bromleybr.co.uk/reflex/](http://bromleybr.co.uk/reflex/) which is based on nearly up-to-date core code.

rubberfashion:
i see .... at your site it looks like that. 
screenshot_20170507-193507
Thank you for your help!!

gburton:
So far as I recall, both the 
shipping options https://github.com/gburton/Responsive-osCommerce/blob/master/checkout_shipping.php#L291
and the
payment options https://github.com/gburton/Responsive-osCommerce/blob/master/checkout_payment.php#L186
have a required attribute.

We should therefore have no .js checking on them at all as the continue button cannot activate if nothing is chosen.

gburton:
Feel free to restart this as an Issue https://github.com/gburton/Responsive-osCommerce/issues

PR #552, opened 2017-05-15
Update express_payflow.php

crazy-colorado:
Misspelling line 232

tep_db_perform(':table_addres_book', $sql_data_array);

should be 

tep_db_perform(':table_address_book', $sql_data_array);


gburton:
TY

PR #556, opened 2017-05-15
Update backup.php

crazy-colorado:
I understand that tep_db_table_name will add ":table_" to these variables; however, for consistency throughout the system, would it not be best to include it here.


gburton:
TY

PR #558, opened 2017-05-15
Update sage_pay_server.php

crazy-colorado:
Missing :table_.  

My apologies for the many updates.  This is the last ones that I found.


gburton:
TY

And thank you for the updates!  It saves me some work/time which is always appreciated!

PR #560, opened 2017-05-19
Update login.php with list-group class

crazy-colorado:
If more than three are enabled, alignment is off.  Adding class "list-group" fixes it.


gburton:
TY

PR #580, opened 2017-09-06
Removes install directory after install

drillsar:


gburton:
Invalid.

Code mirrors (pretty much) the main osCommerce build.
Obviously other than changes necessary to integrate Bootstrap.

PR #585, opened 2017-10-31
Custom Data Attributes

gburton:
Introduces a few HTML5 Data Points.


gburton:
Would someone please make the changes and report back any problems...or success...
Thanks

BrockleyJohn:
I've put it into a test store which behaves just as before. Manual inspection shows the attributes populated for products with & without attributes, in and out of stock.

gburton:
Thank you @BrockleyJohn - could you message me your URL so that I can take a look?

BrockleyJohn:
done on the forum - can't figure it on here!

gburton:
TY @BrockleyJohn appreciated!

PR #587, opened 2017-11-05
missing tag in admin orders.php

BrockleyJohn:
Not earth-shattering. Browsers can work out what's going on anyway. Likely a core bug in any case. Ignore or merge as you like!


gburton:
TY

PR #588, opened 2017-11-14
Pass security check on ext out of the box

BrockleyJohn:
Yes, I know, it's a core change.

But I have to concede that failing one of the security checks out of the box when it's so fixable is pants.

Is there a good reason not to do this?

Tested on existing store and new install

### DIFF
--- a/ext/.htaccess
+++ b/ext/.htaccess
@@ -1,0 +1,6 @@
+# $Id$
+#
+# This is used to restrict access to this folder 
+# no directory listings
+
+Options -Indexes


gburton:
I believe this was left down to the Shopowner...as different servers need different code.  
Have you tested it across a range of servers?

BrockleyJohn:
Fair question - I expect it will only solve for apache (which is all I have tried, although various setups) and otherwise be ignored.. The osc distro does contain other htaccess files so it wouldn't be inconsistent to have this too. I will have a go and see what happens elsewhere and can have a look into a multi-platform approach if you think it worthwhile.

gburton:
You're correct - it would not be inconsistent to have it.

gburton:
TY

PR #595, opened 2018-02-11
Font Awesome v5

gburton:
These commits replace FA v4 with FA v5.


gburton:
The language change needs testing to ensure that Scheme/Microdata (breadcrumb) does not error out...

Moxamint:
1) Isn't 'far' pro version only?
2) Would you consider adding fa-v4-shim.js for v4 compatibility?

gburton:
> Would you consider adding fa-v4-shim.js for v4 compatibility?

We only use a few icons (that need changing), so to perform a global search/replace takes just a few minutes...will list them on a post at forum.

gburton:
https://forums.oscommerce.com/topic/412589-font-awesome-v5/

PR #600, opened 2018-02-19
Html validation errors

BrockleyJohn:
Clear html  5 validation errors:

- replace header tags noscript module with a content header module (corresponding install changes)
- make sure all itemtypes have an itemscope
- lose remaining align attribute
- unbalanced tag on specials in product listing module
- shopping cart box module lists no products in a p tag instead of li


gburton:
Would the new module be better placed as a Navigation Module?

BrockleyJohn:
I did think about that to put it right at the top of the page, but thought that semantically it's content rather than navigation. It's pretty prominent at all sizes - see attached
screenshot 2018-02-21 12 39 50
screenshot 2018-02-21 12 40 10

PR #602, opened 2018-03-18
Reduce database requests by using original class "category_tree"

osc2nuke:
Optimizing base-code, decreasing multiple database requests for the same objects.
Why would request the database for something that is already requested and available to use?


gburton:
TY

PR #605, opened 2018-03-18
Update general.php

ArtcoInc:
Allows the current stock on hand to be displayed in the shopping cart


ArtcoInc:
incomplete request.

PR #606, opened 2018-03-18
Modify general.php and language files to display quantity on hand in Shopping Cart

ArtcoInc:


ArtcoInc:
Modify general.php and language files to display quantity on hand in shopping cart

gburton:
TY.  Will leave this as is for now.

PR #609, opened 2018-03-20
Class usage CM Upcoming Products

osc2nuke:


BrockleyJohn:
@osc2nuke Henry I guess you're testing on windows.

some class files are wrong case eg category not Category
also manufacturer, manufacturers

I can't get sourcetree to pick it up as a change - presumably because it's on windows!

BrockleyJohn:
also something (presumably cm_i_new_products) throws

>      

1054 - Unknown column 'parent_id' in 'field list'

select products_id as id, parent_id, products_quantity as quantity, products_price as price, products_model as model, products_gtin as gtin, products_tax_class_id as tax_class_id, products_weight as weight, products_weight_class as weight_class_id, products_date_available as date_expected, products_date_added as date_added, manufacturers_id, has_children, products_date_available from products where products_id = '28' and products_status = '1'

[TEP STOP]

see [http://bromleybr.co.uk/bsfinal/index.php](http://bromleybr.co.uk/bsfinal/index.php)

osc2nuke:
Yes i also noticed that sourcetree not pickup files starting with capital (these are the original oscommerce V3 namings of the classes)

osc2nuke:
1054 - Unknown column 'parent_id' in 'field list'

There is an sql file

osc2nuke:
"Why are we placing the data into a variable $comments? Is there a reason for that..."
I did it in my first commit for one of the cm's.

I was thinking it is more uniform and keep it all PHP

BrockleyJohn:
I haven't looked for the sql but assume that parent_id is on product. If so that's wrong... or is it supposed to be a canonical parent?

BrockleyJohn:
has_children also makes no sense for a product unless I'm missing something

gburton:
This is for the upcoming all-new attributes system.

I've asked Henry to work on getting the product class as finalised as possible so that we can then test it to destruction and ask for any extras we need.  Once it's then "finalised" we commence on changing the modules to use the class...

BrockleyJohn:
OK - @osc2nuke when it's ready and you have a moment please attach the sql

PR #610, opened 2018-03-21
Implemented Attributes plus general optimizations

osc2nuke:


osc2nuke:
You all will going to like it
tumblr_lw5hmzemst1r8omnqo1_400

osc2nuke:
👌 
Dunno why i cannot preserve selected states.
Works without the enforce state.

PR #618, opened 2018-03-31
extra getContent

BrockleyJohn:
I'd like to see these in there...

### DIFF
--- a/contact_us.php
+++ b/contact_us.php
@@ -81,6 +81,10 @@
 <div class="contentContainer">
   <div class="contentText">
   
+  <div class="row">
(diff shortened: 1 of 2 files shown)


gburton:
What about for content inside the form?

BrockleyJohn:
If you mean contact us - 
I don't remember being asked for extra form content...
The getContent is inside the form but it would also need a hook to process the fields.

If you mean shopping cart I could do with an example that doesn't require the whole thing modularising

BrockleyJohn:
I chucked these in as quick things I've done for clients in the past for extra features with minimal change. Both pages are candidates for full modularisation but not by the end of March.

PR #619, opened 2018-03-31
hook register in application_top.php

raiwa:
auto register hooks

### DIFF
--- a/admin/includes/application_top.php
+++ b/admin/includes/application_top.php
@@ -242,4 +242,5 @@
                        
   require(DIR_FS_CATALOG . 'includes/classes/hooks.php');
   $OSCOM_Hooks = new hooks('admin');
-  
+
+  $OSCOM_Hooks->register(strtok($PHP_SELF, '.'));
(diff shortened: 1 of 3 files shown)


BrockleyJohn:
Doesn't it also need the register taking out of orders.php? Or does registering twice not matter?

raiwa:
@BrockleyJohn you are right, forgot it to add to Git, thanks

gburton:
Is this 100% tested, and confirmed working with zero problems?

gburton:
I think it's cleaner using basename as filenames could potentially have a . in them.  UNlikely, but could.
Thanks

gburton:
Oh, please push to modularity2 branch.  
Master branch is static now.

raiwa:
Yes, replaced with basename.

Tests conditions:

- latest 2.3.4 BS EDGE under PHP 7.2
- error reporting set to "all"

Tests performed:

- admin side: orders.php, removed hooks register, edited several orders all ok, no errors
- shop side: points and rewards add-on installed, removed all hook registers in several files (checkout_payment, confirmation, process, create account)
- several checkouts completed with points and rewards actions, get points, redeem points, create account. no errors, all redeem points actions correct

gburton:
Pulled into m2 branch, thank you @raiwa for the code and the push to m2.

PR #620, opened 2018-03-31
auto load hook language files

raiwa:


gburton:
No plan to change this class.

PR #621, opened 2018-04-01
Register hooks

raiwa:


raiwa:
Tests conditions:

-     latest 2.3.4 BS EDGE under PHP 7.2
-     error reporting set to "all"

Tests performed:

-     admin side: orders.php, removed hooks register, edited several orders all ok, no errors
-     shop side: points and rewards add-on installed, removed all hook registers in several files (checkout_payment, confirmation, process, create account)
-     several checkouts completed with points and rewards actions, get points, redeem points, create account. no errors, all redeem points actions correct

PR #632, opened 2018-04-15
Fix for cart when removing items

Piernas:
Remembered this issue discussed here: https://forums.oscommerce.com/topic/410720-sessions-languages-and-ajax-loading/?do=findComment&comment=1752997
Would be good to add the fix to the final version. It re-calculates the cart when an item is removed to avoid $cart->total and $cart->weight show old values.


gburton:
TY

PR #634, opened 2018-04-17
Breadcrumb Schema

gburton:
HT module for Breadcrumbs + installation. 
Remove older Breadcrumb Code from breadcrumb class.


gburton:
To update;

1. make changes to breadcrumb class file
2. upload new HT Module + Lang File
3. login to admin and turn on

PR #637, opened 2018-04-24
scrambled website welcome

BrockleyJohn:
worse than fingernails down a blackboard...


gburton:
Would be one for individual shopowners.
"on" is correct (for me).

BrockleyJohn:
I've only ever seen on with 'Welcome on board' - I thought it had crept in from a continental cousin where several languages use 'an' for 'to'. To me it's a glaring tyop.

The language evolves all the time so there's no such thing as wrong English any more.

PR #645, opened 2018-05-01
sort out currency in schema

BrockleyJohn:
schema missing currency value in products list


BrockleyJohn:
by which I mean missing a value for currency!

gburton:
TY

PR #648, opened 2018-05-02
action classes used statically so declare methods static

BrockleyJohn:
executes made static... 

also a couple of undefined index notices corrected in update_product


gburton:
TY

PR #649, opened 2018-05-02
paypal express throwing undefined notices in cart

BrockleyJohn:
because checkout flow processing executes even when not enabled


gburton:
TY

PR #652, opened 2018-05-03
add pull down default to state menu

raiwa:
Not a coding bug, but a functional error reported from a real store:

-     customer creates account
-     customer fills in a wrong state name which doesn't match any database state name
-     error is thrown "please select a state from the drop down menu"
-     the first alphabetic state is selected by default
-     customer doesn't pay attention and the first default state is submitted
-     solution: add "Please select" as default to the menu

### DIFF
--- a/create_account.php
+++ b/create_account.php
@@ -414,7 +414,7 @@
         <?php
         if ($process == true) {
           if ($entry_state_has_zones == true) {
-            $zones_array = array();
+            $zones_array = array(array('id' => '', 'text' => PULL_DOWN_DEFAULT));                        
             $zones_query = tep_db_query("select zone_name from " . TABLE_ZONES . " where zone_country_id = '" . (int)$country . "' order by zone_name");
             while ($zones_values = tep_db_fetch_array($zones_query)) {
               $zones_array[] = array('id' => $zones_values['zone_name'], 'text' => $zones_values['zone_name']);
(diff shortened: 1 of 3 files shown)


gburton:
Wouldn't it better to just have the shopowner install a module?  There are at least 5 that cure this behaviour.

BrockleyJohn:
Store owners usually seem to want these pages changing to default to the store country and states but this seems like a good practice change for the core code and it's not going to break anything.

raiwa:
I agree for store country, not so for state.
Remember the state drop down shows only after entering an invalid state name. So the customer had already typed in his state. Why should then show up the store state? Correct is to show "please select" to force the customer to select his correct state. Neither first alphabetic neither store state. IMHO.

gburton:
Yeah, but why can't shopowners just install a module that solves it.   Like I say...there are at least 5, of which I know for a fact that three of them are just HT modules and change no core code.  The other two, no idea if they change core code.  let's get shopowners used to the idea that modules are good and useful and should be used to add in "better".

raiwa:
It's a fix for a functional bug in core. Even with this fix, the available modules are still better. 
As John stated, it dosn't break anything, it doesn't do any harm, but improves core.
Store owners may:
1. not be aware of the problem
2. not know that there are other modules

gburton:
Sorry, I do not see this as a bug.

PR #663, opened 2018-05-07
Fix warning message in admin->modules-> payment -> install module

JAValeryon:
Warning: Use of undefined constant MODULE_PAYMENT_MONEYORDER_PAYTO - assumed 'MODULE_PAYMENT_MONEYORDER_PAYTO' (this will throw an Error in a future version of PHP)

The language file uses the constant MODULE_PAYMENT_MONEYORDER_PAYTO which is not defined until the module is installed.

### DIFF
--- a/includes/languages/english/modules/payment/moneyorder.php
+++ b/includes/languages/english/modules/payment/moneyorder.php
@@ -9,8 +9,14 @@
 
   Released under the GNU General Public License
 */
-
+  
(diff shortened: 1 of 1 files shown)


gburton:
TY

https://github.com/gburton/Responsive-osCommerce/pull/664/commits/5367beea4c758c7513a0f985dfe1330dcd0d7a61

PR #672, opened 2018-05-12
fix continue button to submit button in checkout success

raiwa:
continue button must be form submit button.
see forum messages:
[https://forums.oscommerce.com/topic/398365-purchase-without-account-for-234-and-bs2334/?do=findComment&comment=1770093](url)


BrockleyJohn:
presumably this bug also affects core - eg. product notifications

raiwa:
Yes, that's why I submit the fix

gburton:
TY

PR #673, opened 2018-05-12
Fix missaligned gender form required input

raiwa:
screenshot missaligned:
https://www.oscaddons.com/images/bug_reports_images/required_input_missaligned.jpg

screenshot fix:
https://www.oscaddons.com/images/bug_reports_images/required_input_fixed.jpg

checked also in mobile view and ok


gburton:
There is a reason why the inputs in checkout_new_address are not required.  Cannot remember the reason.
Please remove that file from the push, thanks

raiwa:
DONE

gburton:
TY

PR #674, opened 2018-05-14
separate download module in account and checkout success module

raiwa:
create standarised tpl for checkout success
apply BS styling to modules


gburton:
We are not doing any more feature changes.
TY

raiwa:
Shouldn't at least panel tags be added to the existing download module. It shows extremely unstyled in both places.

PR #704, opened 2019-02-08
Update Broken / Outdated

gburton:
When changing a default currency, ensure to set value on the new default currency to 1, and then update currencies.  We thank https://exchangeratesapi.io for access to their API.


gburton:
Please could someone test this and report back...

PR #723, opened 2019-06-11
fix undefined index in_stock

ruden:


gburton:
Please push to 201906 branch, thanks.

PR #724, opened 2019-06-11
fix undefined index in_stock

ruden:


gburton:
Great thank you @ruden 
This was previously fixed, but then the fix somehow got deleted a bit later.

Anyway, TYVM.  Will merge to master at the end of June along with rest of 201906 changes.

PR #725, opened 2019-06-12
update ht_breadcrumb_schema.php

ruden:


gburton:
Please push to 201906 - thanks.

Is there a problem with $k ?  Why are you increasing it by 1 ?

PR #726, opened 2019-06-12
update ht_breadcrumb_schema.php

ruden:
> Is there a problem with $k ? Why are you increasing it by 1 ?

https://schema.org/BreadcrumbList

> e.g. beginning with '1' for the first item in the list.

and sample code

This is how to count the fingers 1, 2 ... 10

### DIFF
--- a/includes/modules/header_tags/ht_breadcrumb_schema.php
+++ b/includes/modules/header_tags/ht_breadcrumb_schema.php
@@ -33,6 +33,10 @@
       
       $itemlistelement = array();
       foreach($breadcrumb->_trail as $k => $v) {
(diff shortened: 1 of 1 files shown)


gburton:
I like it at zero - it does not produce an error at zero.  
Why are you checking for live link and breaking if not ?

ruden:
Zero is the first key in php array
You think as a programmer
I checked
https://github.com/davejamesmiller/laravel-breadcrumbs/blob/6465e8710341578092fdd8aea24319d4f53ab27c/views/json-ld.php#L14

gburton:
0 or 1, it doesn't matter.  Keep is as zero.

PR #733, opened 2019-06-30
fix constant already defined

ruden:


gburton:
shopping cart will be modularised soon

ruden:
> shopping cart will be modularised soon

This is bad
as well as the modularity of the product page and the index
for width control

PR #743, opened 2019-07-09
Proposal for stylsheet hook

cupidare:
Proposal for stylesheet hooks

I had an issue with the user.css as it was not loaded as the last stylesheet. 
- That's why I want to _propose_ to either separate user.css and custom.css or the insert both of them at the end. In this way I want to ask 
- whether it was possible to _introduce a sort order_ for hooks in order to force the "system" to load them as intended. 
- Additionally when it comes to stylesheets the browser caches it all the time and troubleshooting can be a confusing when forgetting about this. That's why I integrated your version variable in the call directly which one can update when updating the file. In my case Page speed insights detected no change in the file as it was presumably cached already.

### DIFF
--- a/includes/hooks/shop/siteWide/styleSheetUser.php
+++ b/includes/hooks/shop/siteWide/styleSheetUser.php
@@ -1,0 +1,29 @@
+<?php
+/*
+  Copyright (c) 2019, G Burton
+  All rights reserved.
+
+  Redistribution and use in source and binary forms, with or without modification, are permitted provided that the following conditions are met:
+
+  1. Redistributions of source code must retain the above copyright notice, this list of conditions and the following disclaimer.
+
+  2. Redistributions in binary form must reproduce the above copyright notice, this list of conditions and the following disclaimer in the documentation and/or other materials provided with the distribution.
+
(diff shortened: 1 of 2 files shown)


gburton:
From my testing the user.css is always loaded last.  Look at the source code;


<link href="custom.css" rel="stylesheet">
<link href="user.css" rel="stylesheet">

So I don't quite see the issue ?  Can you expand as to why you think the user.css was not loaded last?  Can you add something in (eg custom.css)

body { background: pink }

and then in user.css

body { background: white }

Reload page and body should still be white as user is loaded last.  

Adding the versioning onto the stylesheet name shouldn't make any difference - you'd need to have a constantly changing version to force stylesheet changes in this manner.

cupidare:
Thank you for pointing this out. I installed another package (BS-Select) ... and made a module. Obviously it sneaked before the user.css 
Nevertheless I believe it would be good to have it loaded at the very end of the site (despite some possible performance losses - with page speed insights I am with the fully modified shop still at 98 points with phoenix). As it is the user.css all changes made to it will definetively applied. Or is there a possiblity to specify the sort order of hooks?  

Thank you ever so much.

gburton:
If you want user.css loaded at the end of the site...use the other hook call(s).  See how that is done in (eg) the bootstrap hook.

PR #751, opened 2019-07-18
Update cm_navbar.php

scro0303:
MODULE_CONTENT_NAVBAR_COLLAPSE now has the option of "navbar-expand"


gburton:
IMPORTANT: Please do not create a Pull Request without creating an issue first.

Any change needs to be discussed before proceeding. Failure to do so WILL result in the rejection of the pull request.

Issue #766, opened 2019-08-05
When associated _WAY define is empty, don't add brackets


scottcwilson:
Right now, Flat Rate is presented in the order total line items as 

Flat Rate ():	

since MODULE_SHIPPING_FLAT_TEXT_WAY is empty.  The code should test for this condition and if it's true, not add brackets.

gburton:
I thought that was changed some time back...maybe on a branch then never pulled in.

scottcwilson:
Nope, just updated my repo and it's still there. LMK if you want a PR.

gburton:
Sure thing, PR it.

gburton:
1.0.1.3

PR #768, opened 2019-08-06
Do not display empty parens when WAY is not set

scottcwilson:
Fixes issue #766 

When the _WAY define for the associated shipping method is empty, no braces should be shown. 

Testing:
You MUST go back to the checkout shipping page to test  this change; you cannot simply refresh on the checkout confirmation page. 

Result:

image


gburton:
1.0.1.3

PR #776, opened 2019-08-09
Moves base array for information box into core

### LINKED ISSUE #775: Move box list creation out of language files, allow extension without core change
Implementation of a concept discussed here: 
https://forums.oscommerce.com/topic/493826-adding-to-the-information-sidebox-in-phoenix/?tab=comments#comment-1792609

Allow modders to add entries to boxes (information box in this case) without modifying core files.  Also moves filenames into the core so users wishing to change language files don't see filenames or arrays, just a list of defines.

scottcwilson:
Fixes #775 

Moving the base array out of the language file sets the stage for making it easier to extend. 

**Test plan (required)**
- Tested box with no extension - box has 4 entries as expected. 
Added to box by changing includes/languages/english/modules/boxes/extras/bm_information_example.php to uncomment out the last line to 
```
$information_box_list['better_together_promo.php'] = 'Better Together Offers';
```
in language file.  Box now has 5 entries.

### DIFF
--- a/includes/languages/english/modules/boxes/bm_information.php
+++ b/includes/languages/english/modules/boxes/bm_information.php
@@ -14,8 +14,10 @@
   define('MODULE_BOXES_INFORMATION_DESCRIPTION', 'Show information page links');
   define('MODULE_BOXES_INFORMATION_BOX_TITLE', 'Information');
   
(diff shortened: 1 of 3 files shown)


scottcwilson:
Closing per Gary's direction.

PR #777, opened 2019-08-10
fix escape string

ruden:
https://forums.oscommerce.com/topic/493844-bugs-in-review-intaface-admin-side/


gburton:
IMPORTANT: Please do not create a Pull Request without creating an issue first.

Any change needs to be discussed before proceeding. Failure to do so WILL result in the rejection of the pull request.

PR #803, opened 2019-09-17
add margin-top to back button

peterpil19:
Hi Gary,

Adding mt-2 as a class to the back button creates some much-needed space between it and the and the continue button as they are currently stuck together which is a little unsightly.

I am sure this will be addressed in future when these pages become modularised however makes these pagers appear neater in the meantime.

I have just proposed it for one file in case you don't like the idea to avoid wasting your time. If you do, I can make the changes to all the other relevant files and send through as pull requests.


gburton:
Let's leave this for the future, TY.  Adding a margin here would misalign buttons when in a row.

PR #807, opened 2019-09-27
Category Image index products module

peterpil19:
For consideration - arising out of discussion on the Phoenix Club: earlier versions of BS osCommerce had an option to display category images. This module shows the category image (based on the title module which already had the code for the image, just not in the template file).


gburton:
https://github.com/gburton/CE-Phoenix/blob/79351fa561a48d7754d184ad85d99778cd9381a4/.github/pull_request_template.md

Thanks

PR #831, opened 2019-11-14
Db defined hooks

ecartz:
Create a method to add hooks in the database. This allows hooks to be removed as well as added. And it removes the need to conform with naming conventions.

This will allow Phoenix to ship with certain behaviors preinstalled but allow other behaviors to be substituted by database calls. The current hook system requires that files be deleted to disable them. This makes updates more difficult, as it is hard to tell if a file should be added or if it was deliberately removed. Functionality using this form can simply delete the database reference. So the core hooks can be moved out of the file registration system and use this system instead.

This change is backwardly compatible. Any existing file-based hooks will continue to work. This allows store owners and App developers to deploy hooks as now. But the core code can instead use these instead of file-based hooks. This will leave the hooks directory to the customizers so that updates won't have to change it.

* New hooks table in install/oscommerce.sql

Add a table to define hooks.  This SQL needs to be run manually in updates.  

* Add loading from database in includes/classes/hooks.php

Alters the hooks class so that hooks can be loaded from either files or the database.  Requires the table be created first.

* Move hook code higher in includes/application_top.php

My goal is to make changes as high in the file as possible.  So I needed to move the hook code as early as possible.  Three lines (five with comment and blank) moved.  One added.  None deleted.

* Update copyright year in all three files

Test plan, etc. discussed via PM.  Loose description:  

1.  Download the code and install a store somewhere.  
2.  Verify that the shop displays normally and Bootstrap is loaded.  
3.  Add file includes/legacy/legacy_hook_test.php with content 
```
<?php

  class legacy_hook_test {

    public function test() {
      $GLOBALS['legacy_hook_test_value'] = 'Hello world!';
    }

  }
```

4.  Run SQL on the database:  `INSERT INTO hooks (hooks_path, hooks_site, hooks_group, hooks_action, hooks_code, hooks_class, hooks_method) VALUES ('includes/legacy/legacy_hook_test.php', 'shop', 'siteWide', 'startApplication', '+010', 'legacy_hook_test', 'test') `
5.  Add to some file that produces output, the following code `echo $legacy_hook_test_value;`
6.  Verify that the text "Hello world!" appears where the echo was added.  
7.  Verify that the shop still displays normally.

### DIFF
--- a/includes/application_top.php
+++ b/includes/application_top.php
@@ -5,7 +5,7 @@
   osCommerce, Open Source E-Commerce Solutions
   http://www.oscommerce.com
 
-  Copyright (c) 2014 osCommerce
+  Copyright (c) 2019 osCommerce
 
   Released under the GNU General Public License
 */
@@ -55,6 +55,12 @@
 // make a connection to the database... now
   tep_db_connect() or die('Unable to connect to database server!');
 
+// hooks
+  require('includes/classes/hooks.php');
+  $OSCOM_Hooks = new hooks('shop');
+  $OSCOM_Hooks->register('siteWide');
+  $OSCOM_Hooks->call('siteWide', 'startApplication');
+
 // set the application parameters
   $configuration_query = tep_db_query('select configuration_key as cfgKey, configuration_value as cfgValue from configuration');
   while ($configuration = tep_db_fetch_array($configuration_query)) {
@@ -100,10 +106,6 @@
 // define general functions used application-wide
   require('includes/functions/general.php');
   require('includes/functions/html_output.php');
-
-// hooks
-  require('includes/classes/hooks.php');
-  $OSCOM_Hooks = new hooks('shop');
 
 // set the cookie domain
   $cookie_domain = (($request_type == 'NONSSL') ? HTTP_COOKIE_DOMAIN : HTTPS_COOKIE_DOMAIN);
@@ -408,7 +410,6 @@
     }
   }
 
-  $OSCOM_Hooks->register('siteWide');
   $OSCOM_Hooks->call('siteWide', 'injectAppTop');
 
   $OSCOM_Hooks->register(basename($PHP_SELF, '.php'));
--- a/includes/classes/hooks.php
+++ b/includes/classes/hooks.php
@@ -5,19 +5,78 @@
   osCommerce, Open Source E-Commerce Solutions
   http://www.oscommerce.com
 
-  Copyright (c) 2018 osCommerce
+  Copyright (c) 2019 osCommerce
 
   Released under the GNU General Public License
 */
 
+  function &tep_guarantee_subarray(&$data, $key) {
+    if (!isset($data[$key]) || !is_array($data[$key])) {
+      $data[$key] = [];
+    }
+
+    return $data[$key];
+  }
+
+  function &tep_guarantee_all(&$data, ...$keys) {
+    $current = &$data;
+    foreach ($keys as $key) {
+      $current = &tep_guarantee_subarray($current, $key);
+    }
+
+    return $current;
+  }
+
   class hooks {
-    var $_site;
-    var $_hooks = array();
+
+    public $_site;
+    public $_hooks = [];
+    const PREFIX = 'listen_';
+    private $prefix_length;
 
     function __construct($site) {
       $this->_site = basename($site);
+      $this->prefix_length = strlen(self::PREFIX);
 
       $this->register('global');
+    }
+
+    function sort_hooks() {
+      foreach ( $this->_hooks as &$groups ) {
+        foreach ( $groups as &$actions ) {
+          foreach ( $actions as &$codes ) {
+            uksort($codes, 'strnatcmp');
+          }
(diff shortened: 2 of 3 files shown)


gburton:
Everything seems to work as intended...

I did get one batch of problems;

Warning: call_user_func() expects parameter 1 to be a valid callback, first array member is not a valid class name or object in \1031a\includes\classes\hooks.php on line 122

Warning: call_user_func() expects parameter 1 to be a valid callback, function 'bootStrap' not found or invalid function name in \1031a\includes\classes\hooks.php on line 123
This line repeated for each of the core hooks residing in /includes/hooks/shop/siteWide/

gburton:
I'd also like to move the /legacy/folder into (eg);

/includes/hooks/legacy/shop/
and
/includes/hooks/legacy/admin/

Just to keep all hooks in one place...is there a downside to this?

This may require some attention in the admin/modules_hooks.php page.

ecartz:
> Just to keep all hooks in one place...is there a downside to this?

Just that if someone then came up with a hook object constructed as site legacy/shop or legacy/admin, that it would try to load the database hooks as file hooks.  That was one reason why I was deliberately putting them in a different location.  So all the file-based hooks are in one location and the database hooks are not there.  Because I think it's confusing for some of the files in includes/hooks to be file-based hooks and others to be database-defined.  

I would actually move the existing hooks that ship with core to being database-defined in a new file location.  Then the hooks directory would ship empty and people could add hooks as they wanted.  No site conflicts.  

Also, it's worth noting that the database hooks are deliberately capable of being anywhere (at least as written, anywhere in a class; the class requirement seems removable though) in osCommerce.  So even if we did put the default behavior hook files in includes/hooks, other hooks might not be.  Thus we wouldn't have all the hooks in one place regardless.  

This makes hooks work more like how such things work in other software.  So, for example, a payment module might change behavior somewhere else entirely without having to add an extra file.  And we can change some of this behavior implicitly by configuration.  So someone might say that they want language dependent configuration.  That would change the way that configuration is read from the current method to a new method from the interface without actually changing any code nor setting configurable paths (`if`/`else`) all through the code.  

Beyond those things though, putting the files almost anywhere would work.  

### Admin

> This may require some attention in the admin/modules_hooks.php page.

I had completely forgotten about that page.  But when I look at it, it seems to be surviving.  It does not display any of the database defined hooks, but I don't know that it necessarily should.  I'm not convinced that shop owners should be adjusting the database defined hooks manually.  I suspect it would be trivial to include them in that page, just a matter of adding a query:  
```
      $hooks_query = tep_db_query(sprintf(<<<'EOSQL'
SELECT hooks_group, hooks_path, hooks_action, hooks_code, hooks_class, hooks_method
 FROM hooks
 WHERE hooks_site = '%s'
EOSQL
, tep_db_input(tep_db_prepare_input($file))));
      while ($hook = tep_db_fetch_array($hooks_query)) {
?>
  <tr class="dataTableRow" onmouseover="rowOverEffect(this)" onmouseout="rowOutEffect(this)">
    <td class="dataTableContent"><?php echo $hook['hooks_path']; ?></td>
    <td class="dataTableContent"><?php echo $hook['hooks_action']; ?></td>
  </tr>
<?php
      }
```

### Warning

> Warning: call_user_func() expects parameter 1 to be a valid callback, first array member is not a valid class name or object in \1031a\includes\classes\hooks.php on line 122
> 
> Warning: call_user_func() expects parameter 1 to be a valid callback, function 'bootStrap' not found or invalid function name in \1031a\includes\classes\hooks.php on line 123
This line repeated for each of the core hooks residing in /includes/hooks/shop/siteWide/

Hmm.  That's weird.  I don't get anything like that.  And you say that the site still works?  I would expect that kind of issue to break the site.  

According to GitHub, [122](https://github.com/gburton/CE-Phoenix/blob/a459afc6075373eb5271c1396302befa6ef8a3ae/includes/classes/hooks.php#L122) is a blank line and 123 is 

`     function call($group, $action, $parameters = []) {`

Did you perhaps apply edits to the file rather than downloading a new copy?  Because I see only one line ([127](https://github.com/gburton/CE-Phoenix/blob/a459afc6075373eb5271c1396302befa6ef8a3ae/includes/classes/hooks.php#L127)) on which that warning could appear and you're reporting two different line numbers.  

Maybe try downloading a fresh copy of the file and see what that does?

gburton:
Ok, leave "legacy" as you have it.  I'll try again tomorrow on that other error I came across.  It could very well have been me using it on a "hacked about" test installation...

gburton:
It seems to work, I must have made a copy/paste error.  Have you considered the situation where a Hook file will use two (or more) hooks?  EG:

- https://github.com/gburton/CE-Phoenix/blob/master/includes/hooks/shop/siteWide/bootStrap.php#L23
- https://github.com/gburton/CE-Phoenix/blob/master/includes/hooks/shop/siteWide/bootStrap.php#L30

gburton:
1.0.3.1

gburton:
Pulled 1.0.3.1
TY @ecartz

PR #832, opened 2019-11-14
Refactor to centralize template handling.   (#8)

ecartz:
This centralizes the template handling so that it

* Doesn't have to be repeated everywhere.
*  Can be changed in just a few places and affect everywhere.
*  Prepares for a future where templates can be installed without overriding core code.

I have tested this by viewing a page that uses the new code. I verified that it was using the new code by deliberately breaking it. After fixing the code again, the broken part returned. This can be called like

      $tpl_data = [ 'group' => $this->group, 'file' => __FILE__ ];
      include 'includes/modules/content/cm_template.php';

or

      $tpl_data = [ 'group' => $this->group, 'file' => __FILE__ ];
      include 'includes/modules/block_template.php';

I realize that this isn't much shorter, but that's not the real point. The real advantage here is not related to code length but to reusability. There are basically three patterns that appear about eighty times total: Navbar modules; boxes; and content modules. To change these will take about eighty edits. But this way, once changed they can be modified with just the two template module files.

This change is backwardly compatible, as it doesn't make the old way stop working. It just offers a better alternative.

The main reason that I want this is that I'm going to have to make additional calls like this and would like not to have to redo them later.

* Add template mapping function

Centralize the mapping of files to their templates so as to be able to adjust it later.  In particular, it should be possible for a new template to be installed without overwriting any files that ship with core.

* Create block_template.php

* Create cm_template.php

### DIFF
--- a/includes/classes/osc_template.php
+++ b/includes/classes/osc_template.php
@@ -5,12 +5,13 @@
   osCommerce, Open Source E-Commerce Solutions
   http://www.oscommerce.com
 
-  Copyright (c) 2018 osCommerce
+  Copyright (c) 2019 osCommerce
 
   Released under the GNU General Public License
 */
 
   class oscTemplate {
+
     var $_title;
     var $_blocks = array();
     var $_content = array();
@@ -170,5 +171,10 @@
 
       return $result;
     }
+
+    public function map_to_template($file) {
+      return dirname($file) . '/templates/tpl_' . basename($file);
+    }
+
   }
   
--- a/includes/modules/block_template.php
+++ b/includes/modules/block_template.php
@@ -1,0 +1,16 @@
+<?php
+/*
+  $Id$
+
+  osCommerce, Open Source E-Commerce Solutions
+  http://www.oscommerce.com
+
+  Copyright (c) 2019 osCommerce
+
+  Released under the GNU General Public License
+*/
+
+ob_start();
+include($GLOBALS['oscTemplate']->map_to_template($tpl_data['file']));
+
+$GLOBALS['oscTemplate']->addBlock(ob_get_clean(), $tpl_data['group']);
--- a/includes/modules/content/cm_template.php
+++ b/includes/modules/content/cm_template.php
@@ -1,0 +1,16 @@
+<?php
+/*
+  $Id$
+
+  osCommerce, Open Source E-Commerce Solutions
+  http://www.oscommerce.com
+
+  Copyright (c) 2019 osCommerce
+
+  Released under the GNU General Public License
+*/
+
+  ob_start();
+  include($GLOBALS['oscTemplate']->map_to_template($tpl_data['file']));
+
+  $GLOBALS['oscTemplate']->addContent(ob_get_clean(), $tpl_data['group']);


BrockleyJohn:
I am curious why you pick on this one aspect of all the code duplication in the modules. I now do mine as extensions of a common class that carries most of the drudge and it makes clear what the differences between the modules are.

ecartz:
Because this makes it harder to implement theme switching.  

Originally, if you wanted to install a different design on an osCommerce shop, you had to change the core code in the same files as the business logic.  The current system is better, as the display code is somewhat in different files from the business logic.  But to override it, you still have to edit or replace the files that come with the core code.  Or change the code with the business logic to load different files.  

I want to make it so that a template install could be implemented by dropping in a bunch of files and telling the site to use those files instead of the defaults.  So something needs to change each and every time that a template is included.  If I have to edit eighty core files to implement something, I'd much prefer to do it in such a way that I don't have to do another edit to the same eighty core files next week.  Editing two or three core files is much more practical.  

For most other modules, I don't have an immediate need to make changes to every (for example) payment module.  If I did have such a need, I would be focusing on that kind of change.  For example, I do have a need for checking for conflicts in modules.  So at some future time, expect me to make a proposal to change that in core (by making admin call that code).  But I'm still working on that and may have to change the interface.  I'm much surer of the interface here.  It's the internal functionality that is still in question.  

This shifts things such that future changes to this behavior can be made somewhere in the three files involved here (two new and one edited) once the core files have been edited to make use of the two new files.  

**TL;DR**:  this blocks making a real template system where templates are drop-in additions that can be selected.

BrockleyJohn:
Is theme switching a big thing, then?

Anyway, good to have more opportunity for changing the presentation while modifying/replicating much less core code. 
Can you explain how the theme reroutes the template includes without a core mod?

gburton:
1.0.3.1

ecartz:
Loosely speaking, I expect to implement something like 

```
return isset($template_map[$file]) ? $template_map[$file]
                                   : dirname($file) . '/templates/tpl_' . basename($file);
```

If there is no theme installed, the template map would just be an empty array.  But with a theme installed, the template map could override any or all of the default presentation (except for template_top and _bottom, which would have to be handled separately).  

Another alternative would be to put the template_mapping function inside the theme.  So each theme would have to implement one.  That exact decision can be made later if we have the template loading centralized so that it can be changed in just a few places.

osc2nuke:
The problems with templates results in one simple conclusion.
Oscommerce uses many Content_modules and block_modules.
I wanted "default" fall back for a template file.

This means:
A module has it's own template file.When someone build a template that cover for let's say 95% of the original modules.
The modules that are NOT covered by the offered template, still would work.
As they fall-back to the original oscommerce "Bootstrap" .

Just like Phoenix's attempt is preparing itself for it.
And stated by my years ago.
That each aspect should be named.

For sure ideas expand.
But when i was working on it.
I made it pretty simple:
If the FILE exist (OSC_TEMPLATE . 'tpl_' . basename($file)') //load it......... as a template is set
If NOT................
Load the default template file.( like we do now)

It might could be a discussion if template files should be in the same directory as where the module is located.

For now that not make sense to me if go support templates.

Better to have them in one place like templates/:
- default
- template_1
- template_2

But not in modules/*the module or box/the_file.php
And inside.......
templates directory
That doesn't make sense. ( my point of view)

So the template that came with the module.

How "WE" choose that template is read.............
That is another story.
In the core there is ENOUGH references that would allow to access these "Core" templates by references.
So even if they not excist in the template.
The template itself is able to handle simple css references.

It would not be smart to have a template that would not use these references. ( core stuff).

osc2nuke:
Keep in mind.
Theme builders where never in favor to oscommerce.
For a reason.
Theme Builders manipulated the core to match their view to the theme.
And i , after years experience somehow agreed to this view.
As it underestimates oscommerce core. 

But now the focus is more on core.
So if make theme, NP, yet default modules should still work.
( they might not go look so nice, but here is the theme maker to step in, it will be THEIR responsibility)

osc2nuke:
Let me say one thing...........
I am in 100% favor of have a template choice for oscommerce.

But in this case there are RULES involved.
Allow a template is simple as that.
I even made it so to not even have to use bootstrap as a GUI.
I made it possible to use other frameworks. ( yep........ i went that far).
For now you are stuck with bootstrap.

gburton:
1.0.3.1

PR #835, opened 2019-11-26
Set up application before registering hooks

ecartz:
The siteWide hooks sometimes need the setup from application_top.php when used with third-party Apps.  So return their register invocation to the original place and make the new location use a different group.

**Test plan (required)**

This change has already been tested at https://forums.oscommerce.com/topic/494447-v1040/?tab=comments#comment-1798977

The other way was breaking third-party Apps.


gburton:
1.0.4.1

PR #854, opened 2019-12-08
Catalog autoloader

ecartz:
This change adds an autoloader for the catalog side of the shop.

I have tested it by removing the class includes from application_top.php and navigating a test site (without testing checkout, login, or registration, but some modules already started using it).

My suggestion would be that we don't try to make that change now.  Instead, IMO, we should release this with a build point release (e.g. 1.0.4.1) as is and see if it causes any issues.  We can update application_top.php to not manually include class files after the next patch release (1.0.5.0, so for 1.0.5.1 or later) as part of other changes.  Also we could change the module loading then so that all of them use the autoloader and remove the redundant loading code.  

If you want me to do more testing before this change, let me know.  I'd just have to make a new test site, which is not so difficult.  But my current testing at least shows that it is sane.  

We also might consider making some naming changes to simplify the autoloader logic.  For example, we could move the cc_braintree directory and ht_mailchimp_360 directory to apps alongside PayPal.  We could change the names of alertBlock to alertbox and osC_Actions to actions to match the current filenames.  We could rename camelCase classes to snake_case to match the filenames.  And of course, we could start namespacing classes.

### DIFF
--- a/includes/application_top.php
+++ b/includes/application_top.php
@@ -49,6 +49,10 @@
     define('DIR_WS_CATALOG', DIR_WS_HTTPS_CATALOG);
   }
 
+// autoload classes under the classes or modules directories
+  require 'includes/functions/autoloader.php';
+  spl_autoload_register('tep_autoload_catalog');
+
 // include the database functions
   require('includes/functions/database.php');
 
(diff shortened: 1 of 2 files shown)


gburton:
1.0.4.1
TY

PR #860, opened 2019-12-09
Move autoloader.php to includes/functions

ecartz:
Well, I tested it in includes/functions last time and it worked.  So that should probably be where it goes.


gburton:
TY

PR #862, opened 2019-12-12
Change library domains to Cloudflare

ecartz:
* Move external library requests to Cloudflare

The gtmetrix.com site complains that the demo shop uses more than four domains. This change moves the jQuery, BootStrap, and FontAwesome scripts to the CDNJS Cloudflare CDN. That drops the number of domains on the demo site to exactly four: template.me.uk, cdnjs.cloudflare.com, www.googletagmanager.com, and www.google-analytics.com. The Google site links are not distributed with the core install and probably need to be downloaded directly regardless. So this leaves only the store itself and the CDNJS domain for the core install.  

Updates Bootstrap to 4.4.1 to match the 1.0.4.1 branch of CE-Phoenix.

Note:  this has NOT been tested yet.  For various reasons, I won't be able to test it until tomorrow (possibly this evening for some time zones).  Requesting pull by request.


gburton:
TY.  I'll test this today.  You're right, the two Google Scripts are only what I have installed and do not come in an "out of the box" install.

gburton:
TY

PR #871, opened 2019-12-18
Prepare for hooked includes

ecartz:
Add generate function in hooks class and some general code cleanup.  

Move hook creation higher in admin/includes/application_top.php to match catalog.  Make both use the generate function.  

Move defines and SSL/PHP_SELF configuration after the first hookpoint.  

Eliminate compatibility functions except for setting timezone.  

Remove use of deprecated TABLE_CONFIGURATION.

This is the code that I've been using on my test site.

### DIFF
--- a/admin/includes/application_top.php
+++ b/admin/includes/application_top.php
@@ -27,11 +27,25 @@
     include('includes/configure.php');
   }
 
+// include the list of project database tables
+  require('includes/database_tables.php');
+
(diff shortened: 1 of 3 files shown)


gburton:
Conflict with previous Push.
Will look more in-depth later unless you do it first.

ecartz:
I merged the autoloader code into the hooked-includes branch, so the conflict is gone.

gburton:
1042

PR #873, opened 2019-12-20
Add missing comma

ecartz:
SQL query is missing a comma between the catch-all selections and the single column for the categories_id.  

Tested in forums:  https://forums.oscommerce.com/topic/494521-v1041/?do=findComment&comment=1800386


gburton:
TY

PR #875, opened 2019-12-30
Change to match the language file

ecartz:
In the language file, there is a BUTTON_CONTINUE_CHECKOUT_PROCEDURE but no TITLE_CONTINUE_CHECKOUT_PROCEDURE.  So changing this file to match the language file as well as other files like checkout_shipping.php

I haven't tested this, but it is currently broken and would not be more broken.  

https://github.com/gburton/CE-Phoenix/issues/874

### DIFF
--- a/checkout_payment.php
+++ b/checkout_payment.php
@@ -194,7 +194,7 @@
   ?>
 
   <div class="buttonSet">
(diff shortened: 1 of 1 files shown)


ecartz:
Oops.  Never mind.  Wrong language file.

PR #877, opened 2020-01-02
Move includes/classes/shopping_cart.php to includes/system/versioned

ecartz:
* Rename includes/classes/shopping_cart.php to includes/system/versioned/1.0.0.0/shopping_cart.php

* Autoload the shopping_cart class

Tested by adding a product to the shopping cart on my test site (which currently doesn't have any of the class requires in includes/application_top.php).  

Proposal for addressing the problem raised in https://github.com/gburton/CE-Phoenix/issues/849


gburton:
TY

PR #879, opened 2020-01-06
Remove autoloadable class requires

ecartz:
Most classes can be loaded by the autoloader rather than manually.  This change removes the manual requires.  

Note that osC_Actions cannot be loaded automatically, as it doesn't match the naming scheme.

Tested on my test site this way since the autoloader was added.  Which is why I know that osC_Actions can't be autoloaded under the current setup.


gburton:
TY

PR #880, opened 2020-01-11
Autoloader typo

ecartz:
The snake_cased version will be all lower case. Fixed PasswordHash in catalog and admin.

Tested by adding a product to the shopping cart, which wasn't working but does after this change.


gburton:
TY

PR #888, opened 2020-01-22
Default image

ecartz:
If there is a default image configured and the image source is not a valid file, replace the image source with the default image.  

Add configuration for the default image.  Default to blank so it doesn't do anything.

Tested by renaming a product image and hardcoding a define.  Other images continue to work but this one showed the default.


gburton:
TY

PR #889, opened 2020-01-23
Email needs urlencode

ecartz:
An email can contain symbols that have special meaning in URLs, like plus signs.

https://github.com/gburton/CE-Phoenix/issues/864


gburton:
TY

PR #890, opened 2020-01-23
Db varchar change

mfleeson:

**Test plan (required)**

Hi Gary
Here's the updated master install sql changed to varchar(255) and also an extra script that just modifies an existing installation.
Cheers
Mark


ecartz:
I do not think that we should change currencies' symbols nor hooks.  The currency symbols because I can't see any reason why they'd ever need to be longer.  I mean, those are typically things like $ or EUR.  Hooks because it can break the index on that table and those values are set arbitrarily. It's better to limit them to working values.  The products_gtin is at most 14 characters if present.  For similar reasons, the sesskey does not need to be longer; we set it and don't need it to be 255.  whos_online.time_last_click should also only be one size.  customers_password is another that really is never going to be longer. 

While an update script is helpful, I don't think that it should be submitted into the repository.

mfleeson:
Hi Matt

Having worked on financial trading desks and on international cross currency sites we used a larger field on the currencies symbols to add explanation information; increasing them doesn't mean you have to and doesn't take up database space it just allows for future posibilities.
If you look at the hooks modification I've not increased the index fields to 255 but to an increased amount that still complies with the limit of the combined index.
Sesskey I agree doesn't need to be increased but I was future-proofing for php8 where session keys are likely to increase in size.
whos_online.time_last_click was the one I wasn't sure of because if I was writing / updating it I'd have changed it to a date/timestamp not a varchar.

The update script was purely included so that it could be put into a post here on the forum and that Gary had everything I'd worked on. Personally I'd remove the update from the actual git release but it seemed sensible to offer both sets of code together so the update code didn't get lost.

ecartz:
The index limit is 767 bytes.  Since MySQL's UTF-8 characters can be three bytes long, that gives a maximum index size of 257 characters.  I find that easier to remember as 255 characters to match the VARCHAR limit.  Your version would have them at 555 characters, which would take them to 1665 bytes, more than twice the limit.  

You submitted a pull request that would put the upgrade SQL in the Git release.  In order to make that not happen, we will have to do something.  That something would have to be done outside GitHub, as it doesn't support cherrypicking commits.

mfleeson:
The hooks updates only takes the index to 300 characters i.e. 150 each, so not sure how you've got 555 characters. I normally use github to keep track of solo projects so I apologise if the the upgrade sql is in the wrong place. If you'd all like I can cancel the PR, modify anything you want changing and resubmit.

ecartz:
I was including the code change as well, but you're right that didn't end up in the index (I believe that I made those three add up to something less than 255 because I was planning to put them in the same index, but obviously I didn't do that).  300 is still more than 255 (and 900 is more than 767).  But you're right, 555 is not the issue.  

Let's leave time_last_click as 14 characters.  It would probably be better as a timestamp, but let's not get too far into the weeds on this.  Let's leave sesskey for now.  Let's not monkey with the hooks table.  And I think that products_gtin and customers_password should stay as they are.  

When done, please install the store somewhere.  It would be nice to post a link to it, but I don't know if that's absolutely required.  The point being that some of the changes might cause the install to fail, which would be bad.  So please run it to verify that the changes don't cause problems.  

It would be nice not to have the redundant merge request commit but if it's too much trouble to ditch it, that's not an absolute requirement.  

You might find it easier to wait to see if Gary has any additional comments or concerns before doing too much work.

mfleeson:
The hooks updates only takes the index to 300 characters i.e. 150 each, so not sure how you've got 555 characters. I normally use github to keep track of solo projects so I apologise if the the upgrade sql is in the wrong place. If you'd all like I can cancel the PR, modify anything you want changing and resubmit.

gburton:
I can do cherry-picking at my end, but it is a total PITA to do so and so I prefer not to.

PR #895, opened 2020-01-27
Deprecate tep_class_exists()

ecartz:
Deprecate the tep_class_exists function, which is now just a wrapper for class_exists. Will allow the function to be deleted in some future version.


gburton:
TY

PR #898, opened 2020-01-30
Remove unnecessary array_keys

ecartz:
Fix (copy and paste?) error.  Reported and tested in the forums at https://forums.oscommerce.com/topic/494736-whileeach-update-to-foreach-issue/

I actually think this file could be deleted, since the autoloader will load the catalog version if the admin version is not there.  But it might as well be correct first.


BrockleyJohn:
I suggested removing this when doing the php7 changes as it wasn't used anywhere. IIRC the reason for keeping it was for addons but that would no longer hold true.

gburton:
TY

PR #899, opened 2020-02-01
Update .htaccess for Apache 2.4

drillsar:
This is an update for it to work with Apache 2.4

**Test plan (required)**

### DIFF
--- a/images/.htaccess
+++ b/images/.htaccess
@@ -1,12 +1,11 @@
 # $Id$
-#
+# change example.org to your domain name
 # This is used to restrict access to this folder to anything other
 # than images
 
 # Prevents any script files from being accessed from the images folder
 <FilesMatch "\.(php([0-9]|s)?|s?p?html|cgi|pl|exe)$">
-   Order Deny,Allow
-   Deny from all
+  Require host example.org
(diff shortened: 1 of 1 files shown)


ecartz:
This is incorrect.  There should be no site, not even a shop's own site, that can run .php files in the images directory.  The correct 2.4 (mod_authz_dbm) equivalent of the current configuration is 

```
Require all denied
```

I'm not sure whether we should switch to the 2.4 configuration.  The 2.2 directives still work with mod_access_compat.  The only problem is when mixing 2.2 and 2.4 configuration, one should not use a less restrictive 2.4 configuration inside something with a more restrictive 2.2 configuration.  But no one should be making some objects inside the images directory tree less restricted.  So this seems like a solution in search of a problem.  

You can read more about this here:  https://httpd.apache.org/docs/2.4/upgrading.html

Also, in general we don't take pull requests to master.  Pull requests should be made to the latest version branch.  That's 1.0.4.5 today but may change soon.  

And you're missing a test plan.  Not only don't you link to a working site, you don't describe any testing method whatsoever.  

This is all stuff that can be discussed in the issue.

PR #902, opened 2020-02-02
Refactor email

ecartz:
1.  Remove admin email classes so that the autoloader uses the identical catalog files instead.
2.  Refactor email classes to simplify them and update to PHP 7 standards.
3.  Modify email sending files to use a new object method to add email messages.  This centralizes the HTML/text decision.  

Tested by sending emails to my test accounts.  

This also resolves https://github.com/gburton/CE-Phoenix/issues/901


gburton:
TY

PR #907, opened 2020-02-10
Bootstrapped admin/reviews.php

Denzel76:
Bootstrapped admin/reviews.php based on admin/specials.php.

Insert/Edit review merged together to save code

removed databasetable defines against hardcoded tablenames

updated language file

### DIFF
--- a/admin/includes/languages/english/reviews.php
+++ b/admin/includes/languages/english/reviews.php
@@ -22,7 +22,7 @@ define('ENTRY_PRODUCT', 'Product:');
 define('ENTRY_FROM', 'From:');
 define('ENTRY_DATE', 'Date:');
 define('ENTRY_REVIEW', 'Review:');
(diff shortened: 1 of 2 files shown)


ecartz:
Our process is not always evident.  

We never add to master directly.  We always add to a development release and then push an entire version to master.  So master always represents a release (currently 1.0.5.0).  Development is always done in a branch.  

While the current development branch is 1.0.5.1, we don't want to add more changes to that branch.  We're trying to just do bugfixes to that branch.  When we push it to master, we'll cut a 1.0.5.2 branch.  Then this could be merged there.  

If you want, you could make things easier on us by closing this PR, replacing the tabs with spaces to make the indenting consistent regardless of tab width, and issue a new PR once we've cut the 1.0.5.2 branch.

Denzel76:
Waiting for 1.0.5.2

PR #926, opened 2020-03-01
Separate building notifications from updating stock

ecartz:
This is a preparatory change for adding notification modules.  That in turn is needed to make payment module notifications work properly if they don't include the checkout process language file.


ecartz:
I have the next stage of this at https://github.com/ecartz/CE-Phoenix/tree/notifications

I just have to fix my server's email so that I can test that they work.

PR #928, opened 2020-03-02
Immediate term proposal for error handling

ecartz:
As per an outside discussion, it is necessary for hooks/Apps to have a standard way of blocking form processing.  This would allow any hook/App to call a standard function `tep_block_form_processing()` which then figures out how to block processing for that page.  

Also adds hook points on two pages that did not previously have them for consistency with other pages that do.


gburton:
TY

PR #931, opened 2020-03-06
Update for Apache 2.4 compatible with 2.2

ecartz:
Resolve https://github.com/gburton/CE-Phoenix/issues/900

This retains the existing configuration for sites using Apache 2.2 or older but uses the newer configuration if available (in 2.4 or later).  

Tested by entering the configuration into the httpd.conf file in a 2.4 site.  Not tested in the .htaccess files nor on an Apache 2.2 site.

### DIFF
--- a/admin/backups/.htaccess
+++ b/admin/backups/.htaccess
@@ -19,6 +19,12 @@
 Options -Indexes
 
 <Files *>
-Order Deny,Allow
-Deny from all
+  <IfModule mod_authz_core.c>
+    Require all denied
+  </IfModule>
+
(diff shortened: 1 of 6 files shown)


gburton:
Live (shopside) at https://template.me.uk/phoenix/ if you wanted to take a look...
All seems OK to me.

ecartz:
https://template.me.uk/phoenix/includes/application_top.php returns the correct result (resource forbidden).  And so does https://template.me.uk/phoenix/images/application_top.php

That seems fine to me.

PR #933, opened 2020-03-09
Email from

gburton:
@ecartz


ecartz:
Yes, I'd like to merge this.  But do we want it to go to 1.0.5.4?  Or have we missed that train and need to go to 1.0.5.5?

gburton:
Ah yeah, 1055 now isn't it.  Repush to 1055?

PR #937, opened 2020-03-15
1.0.5.5

gburton:
Pre PR for v1.0.5.5


gburton:
> What do you expect "Hook calls for individual pages" to do?

Seems to work as intended.  To allow people to (eg) hook a javascript only on (say) orders.php
Did you try it ?

ecartz:
> Seems to work as intended. To allow people to (eg) hook a javascript only on (say) orders.php
Did you try it ?

No, but that code would not do that.  Have you tried it without the four new lines of code?  Because that was something that I implemented as part of pipeline hooks.

PR #941, opened 2020-03-22
remove admin cagegories products count

raiwa:

**Test plan (required)**

### DIFF
--- a/admin/categories.php
+++ b/admin/categories.php
@@ -5,7 +5,7 @@
   osCommerce, Open Source E-Commerce Solutions
   http://www.oscommerce.com
 
-  Copyright (c) 2019 osCommerce
+  Copyright (c) 2020 osCommerce
 
   Released under the GNU General Public License
 */
@@ -858,10 +858,7 @@ $('#products_date_available').datepicker({
       if (isset($_GET['search'])) $cPath= $categories['parent_id'];
 
       if ((!isset($_GET['cID']) && !isset($_GET['pID']) || (isset($_GET['cID']) && ($_GET['cID'] == $categories['categories_id']))) && !isset($cInfo) && (substr($action, 0, 3) != 'new')) {
-        $category_childs = array('childs_count' => tep_childs_in_category_count($categories['categories_id']));
-        $category_products = array('products_count' => tep_products_in_category_count($categories['categories_id']));
-
-        $cInfo_array = array_merge($categories, $category_childs, $category_products);
+        $cInfo_array = array_merge($categories);
         $cInfo = new objectInfo($cInfo_array);
       }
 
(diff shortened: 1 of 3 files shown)


ecartz:
Don't we still want to warn when deleting a category with sub-categories and products only in that category?  

Also, when @gburton said "development branch" I thought that he meant 1.0.5.6 rather than gb_dev.  But of course that's something to discuss with him.

raiwa:
If we leave it on delete category action, it will also load slow on root categories.
I believe it is not really needed to know how many subcategories and products will be deleted. Just an alert that the included subcategories/products will also be deleted should be enough. 2 functions can be removed from general.php which are only used for that purpose.
Please discuss and decide with Gary.

Until know Gary always mentioned the exact version where to push, This time I understood therefore the gb-dev branch. Maybe I'm wrong. Please clarify with Gary and let me know. I'll push again then to 1.0.5.6.

ecartz:
> Or do you mean just a more simple check if there is at least one subcategory and/or product contained in the deleted category?

Yes, a simple one or two query check.  If the category has any sub-categories or if it has products (ideally only warning if there is a product only in that category, but the simpler any p2c entries is fine).

gburton:
TY.  Dev branch is always the next version up from the live version.  
Pls push to branch 1056

raiwa:
O.K. I'll try to find time the next couple of days.

ecartz:
```
SELECT COUNT(*) AS total FROM 
(SELECT categories_id AS id FROM categories WHERE parent_id = %1$d
UNION
SELECT p2c.products_id AS id
 FROM products_to_categories p2c
  LEFT JOIN products_to_categories self ON p2c.products_id = self.products_id AND p2c.categories_id != self.categories_id
 WHERE p2c.categories_id = %1$d AND self.categories_id IS NULL ) combined
```
I think this is the correct check query and can be done performantly.  You may want to check the query on your slow site in phpMyAdmin to verify that.  The idea is that this query should return a non-zero result when the category contains either subcategories or products that are in no other category.

ecartz:
Or even 

```
SELECT SUM(combined.type=1) AS category_count, SUM(combined.type=2) AS product_count FROM 
(SELECT 1 AS type, categories_id AS id FROM categories WHERE parent_id = %1$d
UNION
SELECT 2 AS type, p2c.products_id AS id
 FROM products_to_categories p2c
  LEFT JOIN products_to_categories self ON p2c.products_id = self.products_id AND p2c.categories_id != self.categories_id
 WHERE p2c.categories_id = %1$d AND self.categories_id IS NULL ) combined
```
which would return the counts.

raiwa:
Both work nice and fast on all categories except for the deepest category.
The first returns: Array ( [total] => 0 ) 
And the second: Array ( [category_count] => [product_count] => ) 
although the category contains 2 products

raiwa:
Forgot to mention: using `$_GET['cID']`

raiwa:
Complete snippet:
`//        $subcategory_products = array('products_count' => tep_products_in_category_count($categories['categories_id']));
/*      $subcategory_products_check = tep_db_fetch_array(tep_db_query("SELECT COUNT(*) AS total FROM 
(SELECT categories_id AS id FROM categories WHERE parent_id = '" . (int)$_GET['cID'] . "'
UNION
SELECT p2c.products_id AS id
 FROM products_to_categories p2c
  LEFT JOIN products_to_categories self ON p2c.products_id = self.products_id AND p2c.categories_id != self.categories_id
 WHERE p2c.categories_id = '" . (int)$_GET['cID'] . "' AND self.categories_id IS NULL ) combined"));
*/
      $subcategory_products_check = tep_db_fetch_array(tep_db_query("SELECT SUM(combined.type=1) AS category_count, SUM(combined.type=2) AS product_count FROM 
(SELECT 1 AS type, categories_id AS id FROM categories WHERE parent_id = '" . (int)$_GET['cID'] . "'
UNION
SELECT 2 AS type, p2c.products_id AS id
 FROM products_to_categories p2c
  LEFT JOIN products_to_categories self ON p2c.products_id = self.products_id AND p2c.categories_id != self.categories_id
 WHERE p2c.categories_id = '" . (int)$_GET['cID'] . "' AND self.categories_id IS NULL ) combined"));

echo '<br><br>$subcategory_products_check' . print_r($subcategory_products_check);
`

ecartz:
> although the category contains 2 products

But are the two products unique to that category?  If a product is in more than one category, that query won't warn on it.

raiwa:
Yep, missed that point.
I'll prepare and push the first version with a generic message if something is found. It anyway doesn't go deeper then one level and displaying numbers or separate category/product message would confuse.

PR #949, opened 2020-04-07
fix break of table for long version numbers

cupidare:
In order to display the table during install correctly on all viewports


gburton:
TY

PR #955, opened 2020-04-19
Pre PR for 1.0.5.9

gburton:


install/templates/pages/index.php
```diff
@@ -48,8 +48,16 @@
...
-      $warning_array['php_version'] = 'The version of PHP must be at least 7.  The version here is ' . PHP_VERSION;
+    $php_version_thumb = '<i class="fas fa-thumbs-up text-success"></i>';
+
+    if (version_compare(PHP_VERSION, PHP_VERSION_MIN, '<')) {
+      $warning_array['php_version'] = sprintf('The version of PHP must be at least <strong>%s</strong>.  The version here is %s.', PHP_VERSION_MIN, PHP_VERSION);
+      $php_version_thumb = '<i class="fas fa-thumbs-down text-danger"></i>';
+    }
+    
+    if (version_compare(PHP_VERSION, PHP_VERSION_MAX, '>=')) {
+      $warning_array['php_version'] = sprintf('The version of PHP must be lower than <strong>%s</strong>.  The version here is %s.', PHP_VERSION_MAX, PHP_VERSION);
```
ecartz, line 59: I don't think the wording is right here.  It's not "must be".  It's "Performance on versions higher than has not been tested."  We have not received a single valid error report on PHP 7.4.  Valid meaning that there was an actual logic bug that produced the wrong result.

gburton: Updated TY

PR #1035, opened 2021-01-19
Only clear radiobuttons on selection rows when selection changes

gburton:
https://github.com/gburton/CE-Phoenix/issues/1034


gburton:
TY @BrockleyJohn

